PR #1, opened 2022-04-07
Remove non-existent provider from META-INF/services

carlosame:
org.apache.bsf.BSFManager is not available from this Jar file, which is causing issues with modular JDKs. See for example:

- https://issues.apache.org/jira/browse/XALANJ-2632
- https://marc.info/?l=xalan-dev&m=164935208630930&w=2
- https://stackoverflow.com/questions/54682417/unable-to-derive-module-descriptor-provider-class-x-not-in-module
- https://issues.redhat.com/browse/WFLY-10279
- http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-April/012147.html

And the workarounds aren't nice: https://github.com/moditect/moditect-gradle-plugin/issues/12#issuecomment-637543434

Please fix this.


jkesselm:
Xalan-Java does reference BSF during compilation, because ExtensionHandlerGeneral.java has the ability to use the BSF API to run scripts. But I agree that we should not be saying we provide BSF; folks who want to use it should fetch the current release of bsf.jar separately.

This change looks reasonable to me. Unfortunately I do not have write access and so can't accept and merge the pull request.

carlosame:
> I suggest reconsidering and merging the suggestion instead. It would result in showing carlosame among the contributors (e.g. https://github.com/apache/xalan-java/graphs/contributors ), and it would encourage the others.

Thank you for your comment, however yesterday after this PR was closed I removed my Xalan fork and I'd rather not redo the work.

Moreover I have no interest in having my name associated with Apache Xalan. I watched how 2.7.3 directly included the BCEL packages despite my warn about split packages (modular JDKs will reject Xalan if BCEL happens to be also in the modulepath). Then how your PR was ignored "because we do not need CI now" despite it offering other improvements, and then for example a more quirky alternative for test execution being merged instead. Etc... you just have to read the mailing list discussions.

I do not think that giving negative opinions on other open source projects is desirable, however I'm puzzled to see how the ASF keeps pushing Github as a way to get more contributions to their projects but then it continues with "The Apache Way" as described above. I wonder how many other people are going to get their time wasted due to this.

jkesselm:
Apologies if I offended. I'm just getting back into Apache, and I'm used to working in company git environments where people are less concerned about individual credit than about whether the improvement gets made, so I may be a bit fast on the gun in trying to simplify our backlog, and in closing before consensus is reached. Objection  accepted, I'll try to be more careful about that.



--

() Plaintext Ribbon Campaign
/\ Stamp out HTML mail!

PR #5, opened 2023-06-13
Revert default indent-amount

jkesselm:
See comments in file. Base behavior changes need to consider existing users, and need to be reflected in the test suite. (Ideally *any* behavior change should come with an appropriate regression test...)


mukulga:
I agree with the intent of this XalanJ PR. I've just been able to merge this PR.

PR #7, opened 2023-06-20
chore: renormalize line endings, and add .gitattributes to prevent similar issues in the future

vlsi:
See https://stackoverflow.com/a/47580886/1261287

I executed `git add --renormalize .`, and all the changes are triggered by the normalization.

Note: after `.gitattributes` is merged, the line endings will be automatically normalized in the future, so this is one-time-only PR.


jkesselm:
Makes sense to me.

PR #8, opened 2023-06-20
Add more exclusion patterns to .gitignore

vlsi:
Xalan build generates some files under `/xdocs/style` and similar folders, so it makes sense to exclude them to avoid committing them accidentally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,33 @@
 /bin/
 /build/
+/classes/
+/.idea/
+*.iml
+.DS_Store
+
+# We download the dependencies if they are missing
+/lib/
+/tools/java_cup.jar
+/tmp/
+
+# Generated sources
+/src/org/apache/xalan/processor/XSLProcessorVersion.java
+/src/org/apache/xalan/xsltc/compiler/XPathLexer.java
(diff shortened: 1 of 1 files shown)


jkesselm:
This shouldl be resolved as part of the Maven migration, since output will be into subdirectories of the target/ folders.

vlsi:
@jkesselm , I beg your pardon, yet I can't understand why do you connect this PR with Maven migration. `.gitignore` does not depend on Maven, and Maven does not depend on `.gitignore`. 

Apparently, the improved gitignore could have prevented issues like https://github.com/apache/xalan-java/commit/f51fdd9e72b596ddba4894257f1990846be2fe63

jkesselm:
Just a matter of priority in the queue.

--

() Plaintext Ribbon Campaign
/\ Stamp out HTML mail!

jkesselm:
> @jkesselm , I beg your pardon, yet I can't understand why do you connect this PR with Maven migration. `.gitignore` does not depend on Maven, and Maven does not depend on `.gitignore`.

The Maven migration is higher priority. And after the Maven migration, the xdocs output will be going into the target/ directory, which will already be in the .gitignore so no additional work will be needed.

PR #12, opened 2023-06-27
Fix dyn:evaluate error reporting

jkesselm:
**select="dyn:evaluate('document(foo)')"** has been returning empty-set since dyn:evaluate is only aware of XPath core functions, not XSLT additions or extensions, and **document()** was added by XSLT. We may want to support the additional functions in the future, but adding that would be messy.

In any case, responding to a syntax error by returning the empty set is not good behavior for an XPath interpreter.

Better to throw an exception and ask the user to fix the stylesheet, methinks.

(Yes, a negative test should be added for this. We don't currently have an **exslt-fail** bucket; we probably should, and should have an appropriate spread of tests therein. If folks want to hold up this fix until I've added that and a suitable test, that's justifiable.)


jkesselm:
Annoyingly, throwing an exception may actually not be an appropriate change. Per https://exslt.github.io/dyn/functions/evaluate/index.html

> If the expression string passed as the second argument is an invalid XPath expression (including an empty string), this function returns an empty node set. 

If we accept that as the specification (now that https://exslt.org is dead), the best we can do may be to issue a nonfatal error.

Lemme rewrite this.

PR #78, opened 2023-09-07
Xalan 2.7.3

epilogic-ch:
This is a PL that patches surrogate handling for attributes and text. It is partially based on https://github.com/apache/xalan-j/pull/4#issuecomment-431864049 fix


epilogic-ch:
pulled on wrong branch

PR #83, opened 2023-09-13
DOM2SAX.parse(InputSource) should call DOM2SAX.parse()

dubinsky:
For DOM that is not a `Document`, `DOM2SAX.parse()`, in addition to calling `parse(Node)`, calls the `ContentHandler`'s `startDocument` and `endDocument`, since `parse()` won't call them in this case; `parse(InputSource unused)` just calls `parse(Node)`, so the `ContentHandler` does not gereceive the `startDocument` and `endDocument` events, which are mandated by the SAX specification.

This can cause problems; see, for example, https://github.com/scala/scala-xml/discussions/694

The simplest way to fix this is to change `parse(InputSource unused)` to call `parse()` instead of `parse(Node)` - and this is what this pull request does.


jkesselm:
Sorry for the delay in processing this. With all the changes made since then, the simplest and safest way to apply this seemed to be to cherry-pick the one file you altered rather than updating/reconciling/merging.  I've done so. As I understand it, Git *should* have retained your credit on the change.

Good catch, simple solution. Thanks!

PR #101, opened 2023-10-14
PROPOSED: Cutover from Ant-based build to Maven-based build

jkesselm:
Please review this carefully, and test thoroughly.  I have been figuring out Maven as I went, and I'm sure some parts of this are not idiomatic, or not as elegant as they should be.  CAVEAT: Some things wind up in different places than they did in the Ant build. If that bothers folks we can do more work to resolve it, but I think what I haven't already fixed is mostly harmless.

SUMMARY OF CHANGES:

Restructures serializer, xalan, and samples as sub-modules with dependencies on each other. Adds packages to the samples, and fixes some javadoc (see below), but otherwise the actual Java logic should be unaltered. JavaCupRedirect is no longer needed; we might want to review for other dead code.

Avoids carrying binaries in the project; fetch from Maven Central, or (in the case of the Javadoc taglet) carry the source in our own project. As part of that, eliminates dependencies on outdated versions of some of these, specifically java_cup and jlex. JFlex has replaced JLex, which permitted/required handling the look-ahead cases in the grammer rather than by reading ahead via code.

Reworks "site" (documentation) generation. Some files may be landing in diffrent locations in the resulting directory tree; if that's an issue, it can be fine-tuned. Some javadoc errors were (sloppily) fixed before discovering that I can turn off "doclint" and let them continue to be warnings rather than build errors; these should be fixed properly at some point (in many cases by switching to using javadoc inheritance). Some files were converted from HTML to XHTML, since Maven's doc plugin knows how to handle the latter.

On Windows, javadoc needed javax.rmi as an explicit dependency.

TODO: We still have some StyleBook files as well, mostly for the design documentation. I'm handling those with a "stylebook_docgen" script for now, invoked from the mvnbuild script/batch file; that should be moved into the pom's logic.

NOT YET DONE: More of the non-Java files may want to be moved into resource subdirectories.

NOT YET DONE: xalan-test should probably be moved into the Xalan project to facilitate testing during continuous integration, especially if it's no longer in use outside Xalan. First thing to do would be to mavenize that, which would probably require reworking the test drivers from Ant plugins to something Maven-compatible.


jkesselm:
(I'm pretty sure version numbers are not being handled as cleanly/centrally as they should be. That's probably a Maven-idiom issue.)

pom.xml
```diff
@@ -0,0 +1,194 @@
...
+       
+       xalan-test compatibility/success/submodule
+
+       NOTE: To get dependency tree from a multi-module project, use the 
+       command "mvn compile dependency:tree" so everything is in scope.
+  -->
+  
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>apache-xalan-j-2.7.3-maint-main</groupId>
+  <artifactId>apache-xalan-j-2.7.3-maint-main</artifactId>
```
chabala, line 17: This is probably not what you want for your parent coordinates. Considering your existing artifacts at [xalan:xalan](https://search.maven.org/artifact/xalan/xalan/2.7.3/jar), your groupId should probably stay `xalan`, for everything, and neither should embed a version, it has a separate version field. I would suggest artifactId be `xalan-parent` here.

jkesselm: Thanks for the suggestion. I wasn't sure how to handle that, given that what I've made sub-modules was just subdirectories broken out into separate jar files by the ant build script.

Not sure how best to handle the version number so it gets into all the right file names and fields. It looks like Maven generates a Version class for the latter, which we could wrap our existing Version interface around...?

chabala: > It looks like Maven generates a Version class for the latter, which we could wrap our existing Version interface around...?

I'm not sure what this means. Is the existing Xalan build generating some Version interface to make the artifact version discoverable programmatically?

I can tell you that Maven artifacts are versioned, and a parent pom is versioned like an independent artifact. (One could build and publish a parent pom in a separate project, for instance, and then refer to it from many projects.) But, if you set 2.7.3 as the version on the parent pom, and don't specify different artifact versions in the module poms, they will all use the parent pom version at build time. Multi-module builds can be complex and tricky, so some reading is recommended: https://maven.apache.org/guides/mini/guide-multiple-modules.html

jkesselm: I think I'd favor xalan-project over xalan-parent. The intent really is that these really are owned and maintained by a single Apache project/team.

I'm also wondering whether the group ID should really be xalan as it has been, versus apache-xalan or even apache. I don't know what the convention has been for other Apache projects, but as I understand the Maven semantics group ID should, ideally, reflect the group responsible for the package's development and maintenance. Yes,  Maven Central has a field for that, and the Java package names also communicate that, but this still seems the right way to organize the Maven packaging namespace. (xalan:xalan seems redundant, which helps push me in this direction.) Am I misunderstanding Maven intent, or Maven practice?

chabala: > I think I'd favor xalan-project over xalan-parent.

Okay. `*-parent` is just frequently used, and I'd say more idiomatic for parent poms. For example: https://github.com/apache/maven-resolver/blob/master/pom.xml#L25

> I'm also wondering whether the group ID should really be xalan as it has been, versus org.apache.xalan or even org.apache.

Since xalan already has artifacts published at Maven Central, I'd suggest maintaining the existing coordinates. If you do change the groupId, it would be more conventional to follow the reverse domain pattern: https://central.sonatype.org/publish/requirements/coordinates/

But xalan has essentially been grandfathered in with an old groupId, from when folks were just using their project name as a group, e.g. [junit:junit](https://search.maven.org/artifact/junit/junit/4.13.2/jar). If you change it now, you'll need to do more work to make sure Maven knows both sets of coordinates are the same artifact so dependency resolution will work for downstream users.

jkesselm: OK. Fixed the IDs. 

Also fixed some (x)html which was using uppercase tagnames, which HTML may have treated as identical to lowercase but XHTML doesn't.

Thanks for the catch. Any other errors, functional or stylistic?

mukulga:
@jkesselm, I find this pull request (PR) created by you, to merge changes to xalan-java repos's 'master' branch ok.

I think, it'll also be helpful, if you may please also update the XalanJ's README file located at https://github.com/apache/xalan-java/blob/master/README, so that this README file reflects the XalanJ's Maven build process instead of an earlier Ant based build process. You might do this, as part of this PR or as a README file change within subsequent PR(s).

META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,6 @@
+Manifest-Version: 1.0
+Archiver-Version: Plexus Archiver
+Built-By: keshlam
+Created-By: Apache Maven 3.6.3
+Build-Jdk: 1.8.0_382
```
stanio, line 5: Is this a local artifact checked in by mistake, or?

jkesselm:
Good catch re the readme. There are other obsolete docs but let's catch this one.

--

() Plaintext Ribbon Campaign
/\ Stamp out HTML mail!

jkesselm:
Well, grump.

Discovered that xalan-test apparently has a bunch of dependencies on stuff in xalan-java/lib. Eliminating those binaries was a large part of the reason for moving to Maven in the first place. My testing's been working fine because I've been **overwriting** a "normal" xalan-java with the jarfiles from xalan-java-mvn, so those dependencies were still being met.  When I **replace** xalan-java with xalan-java-mvn, the missing libs cause smoketest to fail.

Cleanest stopgap solution I see would be to copy the old xalan-java/lib into xalan-test/lib and fix the test's paths to look there. That should mostly be a no-op for testing pre-Maven Xalan, outside of the risk of the libraries getting out of synch -- but since we don't intend to allow the duplication to last very long, I think it's acceptable.

**PLEASE OBJECT LOUDLY AND SUGGEST ALTERNATIVES** if you think that's a bad idea and have a better solution (short of mavenizing xalan-test, which I'm trying not to do just yet. Otherwise I'll move ahead with that, get it committed, then come back to this changeset.

(We're all committed here. Or should be.)

jkesselm:
Fix applied to xalan-test. Yeah, I cheated and did it on the main. Sorry.

jkesselm:
> Good catch re the readme. There are other obsolete docs but let's catch this one.

New README checked into the maven changeset. Comments welcome.

_"Please be patient. We are making mistakes as efficiently as possible."_

stanio:
@jkesselm, if you haven't saved your previous master try restoring it to a feature branch like:

```sh
git checkout -b maven-build defafee1d5c5  # It happen I have a copy of your remote locally
```

Unless automatic git-gc has kicked in already (shouldn't be that soon) it should succeed.  This is a good example of why one should always create a (separate) feature branch for working on different topics, and not using `master` even in their own fork.

PR #110, opened 2023-10-24
Maven migration, refactored, plus maven build changes since then.

jkesselm:
For history of the previous version of this changeset, see https://github.com/apache/xalan-java/pull/105. This branch's history was restructured, addressing some of the gripes expressed in that one.

Passing our core regression test suite: smoketest, apitest, and (with known non-regression issues) conf.xsltc. Deeper regression testing would not be a bad thing. I'd suggest particular focus on XSLTC, since we have changed the XPath tokenizer and grammar layers. There may also be other edge cases as a result of moving to current versions of libraries rather than the old ones we were carrying around in the Ant build.

Note that there are dependencies on where the serializer resources are located. If someone really thinks they need to be moved, I suggest opening a Jira ticket; there is no need to resolve that now.

This version has better handling for creating the build/ convenience directory (copy jarfiles to where the ant build put them) for compatibility with xalan-test.

This PR does *NOT* yet bring tests inboard. That will be dealt with when we mavenize xalan-test.

**Note:** I think I'm beginning to agree with the folks in the "squash considered harmful" camp. There was a specific reason for using it in previous environment I worked with, but not here. I'll bring that up for discussion on the mailinglist.


jkesselm:
I'll be mostly away from keyboard until Tuesday, visiting friends. It'd be great to have feedback -- or a merge?! -- by then.

Note: I'm not (yet?) generating all the ancilliary distribution zipfiles that the Ant build was. If we need 'em, that can be added; opinions definitely sought.

jkesselm:
In the absense of review: 

Given that the executable build appears to be correct, I'm looking at subtler differences betweeen the Maven and Ant `build/` output

1) In the jarfile for the samples
- Jarfile is renamed from xalansamples.jar to samples.jar. I think that's OK; could be persuaded to change it.
- Samples are moved into packages. That's deliberate, but their documentation and invocation (if any?) should be updated to reflect that.
- Maven version of samples.jar includes clases that the Ant build put in separate sample jarfiles. We could consider producing those targeted archives:
* Ant xsltcapplet.jar: TransformApplet
* Ant xsltcbrazil;jar: TransformerHandler
* Ant xsltcejb.jar:  TransformBean, TransformHome, TransformRemote, TransformServlet
* Ant xsltcservlet.jar: CompileServlet, TransformServlet
* Ant xalanservlet.war:  ApplyXSLT, ApplyXSLTException, ApplyXSLTListener, ApplyXSLTProperties, DefaultApplyXSLTProperties, OrderedProps, PIA, SimpleXSLTServlet, UseStylesheetParamServlet, XSLTServletWithParams. This web-archive should also include the media.properties file,, a copy of xalan.jar, and the xml, xsl, and jsp files to run that web archive.  Note that the jsp file probably needs to be updated to reflect new servlet/ package.
 
Should I fix the Maven build to produce these isolated XSLTC sample jar/war files, or is nobody ever really going to care? (Arguably I should produce the .war file, at least, since it needs data files/directory structure not included in the overall samples jarfile.)

jkesselm:
Another current difference between the ant and maven builds is how the "distribution" archives are created.

Currently, the ant build generates both zip and .tar.gz versions of  xalan-j_version_number-bin.tar and  xalan-j_version_number-src.tar. 

The Ant source collection puts everything under  a top-level xalan-j_version_number/ directory, and pulls xalan-test in as a subdirectory. The intent appears to be to facilitate atomic distribution, build, and test when fetch from git is not practical. Maven instead is currently producing a .sources.jar for each module, and is not including the tests in any of them (though it will do so if/when we migrate xalan-test into xalan_java/test). Is there still a need for a source distribution collection?
 
 The Ant bin collection does *not* include a copy of the tests, but does bundle xalan and serializer under a xalan-j_version_number/ directory, and includes documentation and SOURCE of the samples -- in other words, it's basically the build output minus compilation of the samples. It does *not* bundle the tests. Maven is not currently producing an equivalent. Is there still a need for this?

jkesselm:
Maven build now copies *-sources.jar into top level build/, emulating ant build's gathering of (unified) xalan*-src.zip and xalan*-src.tar.gz.

Other differences noted above persist: tests are not folded into this set of sources, single samples.jar (rename xalansamples?), no xalanservlet.war.

Rechecking jarfile contents to make sure everything still matches outside of those quibbles. Still seeking consensus from the rest of Mission Control.

jkesselm:
STATUS UPDATE:

1) Since we're now compiling the taglet ourselves, javadoc can't run until after that has happened. Changed the command line; would prefer to have dependencies drive that... Best thought I have is to move the javadoc operation into a lightweight module with dependency upon the taglet, possibly into the same one that is now doing the binary zipfile. Is there a better way for Maven to handle that? Note that this needs to happen before bin.zip is generated.

2) Still need to move/copy ./buildBinaryRelease/target/xalan-release-2.7.3-bin.zip and ./target/xalan-project-2.7.3-source-release.zip into the build/ directory for Ant output emulation. (Which I still want to have if we can do it.) Again, there's the timing question of doing the copy only after the bin.zip exists; is that best handled through phase (pre-install?), another dependent module, or will just putting it in package do the right thing.

3) I'm not certain which phase the source-release.zip should be generated during. Arguably, it can occur at any time, since it shouldn't be relying on generated artifacts, so leaving it in packaging of the parent module should work... but may not be stylistically correct. It can be copied to the appropriate place during package, assuming it has been generated by then.

4) I'm still not generating the individual sample jarfiles, or the warfile for the one sample packaged that way. It isn't clear the former separations are needed now that everyone has lots of memory and bandwidth to play with, but the latter is an outstanding issue. Should this be a sub-submodule under samples, or is there a better way to handle it?

jkesselm:
I REALLY want to get this in before other stuff starts arriving and I need to do manual merges. It may be close enough to justify doing so; it *does* produce correct artifacts for almost everything. I wanted Gary to sanity check me before pulling that trigger, but...

jkesselm:
A PR has been issued against xalan-test which should make it compatible with this code. Merging that should cause CI to succeed here.

Suggestion: This might be a great time to move development from the "master" fork to "main", and to snapshot last-ant-build-before-maven-migration (better name welcome).

jkesselm:
Merged the fixes to make Xalan-test work with this build, and ran the CI just as initial probe. 

We need to alter this branch's CI to ensure it is indeed running the Maven build for xalan-javs rather than trying to run ant. 

We may also need to check and fix the CI definition to ensure operations are performed in dependency order, since the run executed in an order I didn't expect.

Investigating. Input welcome.

jkesselm:
Continuous integration YAML has to be adapted to mvn build. I've checked in a proposed change for testing.

jkesselm:
Another change: Reintroduced the doclet. We still can't compile the silly thing, but javadoc seems to run against it anyway as legacy code... so I'm providing decompiled source STRICTLY as documentation, and the jarfile as a legacy blob; it isn't _good_ but it's better.

xalan-test may also want to use this. Or may already be using the blob. Or something.

We may still need to figure out how to do a single site build across both xalan and xalan-test, if we want to fully replicate the existing site. In Ant, we temporarily checked out xalan-test as a child of xalan-java so we could run javadoc across everything contained in both. There is probably a better solution.

jkesselm:
- Factored out samples which should produce their own jar or war files.
- Doing final(ish) jar/war/tar content checks.
- Then will need to reconcile for merge.

CAVEAT: Sample classes originally were in the default (ie, no) package. I've taken the liberty of putting at least some of them under a package reflecting their intent -- general samples, xsltcservlet. There may be outdated references in the sample doc, or html, or jsp files which I haven't reconciled yet. Given how long it has been since anyone looked at these samples, that may not be an issue... but it should be reviewed, and arguably the samples should all be regression-tested (I doubt any of them are).

jkesselm:
Content lists of all Maven build .jar and .war files confirmed as matching Ant equivalents, modulo java packages of samples, how version is handled, and the now-shaded (and newer) java_cup/runtime package.

Todo: 

1. Re-recheck tar/zip file content to make sure nothing shook loose.
2. Then reconcile for merge.

jkesselm:
I'm now getting a reasonable-looking xalan-java src tarball. (About the right number of lines; haven't groveled through all the reorgs yet.)  **_But..._**

In the past, the src tarball/zipfile has included an embedded copy of xalan-test, for the convenience of anyone downloading source who wanted to confirm that their build had succeeded. (The binary distro did *not* include xalan-test.)

Should the maven src distro:

- Include the embedded xalan-test, despite that still being an ant build?
- Make xalan-test produce its own source distro?
- Something else?

(I'd rather not recreate the embedding of xalan-test's source, but it wouldn't be impossible, just ugly.)

jkesselm:
Reconciled with Master.

jkesselm:
Happy New Year. 

Have fun. Fix bugs. Support users. Clean code. Add features. Mostly in that order... ;-)

PR #113, opened 2023-11-05
Resolve Memory Leak

martinhickson:
- Cleared the reader before marking it unused

Please consider either merging this pull request or applying the memory leak fix manually.

### DIFF
--- a/src/org/apache/xml/utils/XMLReaderManager.java
+++ b/src/org/apache/xml/utils/XMLReaderManager.java
@@ -152,10 +152,15 @@ public class XMLReaderManager {
      * @param reader The XMLReader that's being released.
      */
     public synchronized void releaseXMLReader(XMLReader reader) {
-        // If the reader that's being released is the cached reader
-        // for this thread, remove it from the m_isUse list.
-        if (m_readers.get() == reader && reader != null) {
-            m_inUse.remove(reader);
+         if (reader == null) {
+            return;
         }
+        // If the reader that's being released is the cached reader
+        // for this thread, mark it as no longer being in use.
+        if (m_readers.get() == reader) {
+           m_readers.set(null);
+           m_inUse.put(reader, Boolean.FALSE);
+        }
+        m_inUse.remove(reader);
     }
 }


jkesselm:
First time I've seen that "workflow awaiting approval" warning. Any idea what causes it? Side effect of having added Vlad's CI?

Outside of that, the change appears reasonable at first glance.

jkesselm:
For everyone's info: The "1 workflow awaiting approval" message is an artifact of adding Vlad's CI to the Xalan package. Git was asking us us to inspect the code before letting CI run, to ensure against someone having inserted code what will cause trouble when executed in the CI environment.  Not wholly unreasonable, just a new step in our PR evaluation process.

Having done that -- The change looks reasonable and CI (admittedly incomplete) didn't report any regressions, so I'm going to go ahead and merge.

vlsi:
Can the PR have some tests to ensure the leak is indeed fixed, and that the performance was not reduced?

vlsi:
> If you happen to know of a trustworthy implementation that could be used from a regression test, I'd appreciate a pointer

* https://github.com/mjiderhamn/classloader-leak-prevention/tree/master/classloader-leak-test-framework can ensure classes can be unloaded
* https://github.com/openjdk/jol can inspect into object contents. For instance, `GraphLayout.parseInstance(...)`: https://github.com/openjdk/jol/blob/9d62009e97a1756ae91a16ab9e72f294efce17a9/jol-samples/src/main/java/org/openjdk/jol/samples/JOLSample_24_Arrays.java#L75C49-L75C49 or use java heap dump as an input. One can collect heap dump with `com.sun.management.HotSpotDiagnosticMXBean#dumpHeap`

> Especially in hotspot JIT compilers, which can optimize differently every run. Again, specific proposals for reliable automatable metrics that we could add to our toolkit are welcome

https://github.com/openjdk/jmh is the way to go.

For instance, see https://github.com/pgjdbc/pgjdbc/blob/b49b61806f050ed13a255ad35449774297fad8c2/benchmarks/src/jmh/java/org/postgresql/benchmark/escaping/EscapeProcessing.java#L49-L53

> it's just nulling one field

I might be wrong here, however, as `m_readers` becomes `null`, the subsequent calls to `getXMLReader()` become slower since they would have to create a brand new reader instead of reusing the previously cached one in the `ThreadLocal`.

In other words, I believe this PR makes `ThreadLocal m_readers` useless. If the caching was not needed in the first place, then `ThreadLocal m_readers` should be removed. If the caching was needed, then this PR introduces performance regression.
Of course, there's an option I am wrong, however, well, I am quite sure my analysis is right.

jkesselm:
I'll take a look. I know there are a few places where we optimized things so far from inner loop that the optimization really didn't buy us much. On the other hand some caches are critical to performance, and some (like the DTM cache) also have semantic implications.

martinhickson:
Perhaps this cache could be configurable, since it may lead to high memory consumption with a large number of threads, but in other cases, CPU usage optimization may be the priority.

jkesselm:
The biggest reason we need the cache, ***if *** I remember correctly, is semantics rather than performance. XPath/XSLTC make some assertions about node identity that need to be true across multiple references to the document, and that becomes hard to achieve if the document() call retrieves a new set of node objects each time. But that behavior only needs to be true across one stylesheet's execution. We don't want to drop things from the cache while in use, but after all references to them have gone away they can be uncached and GC'd (at the cost of possibly reloading them later.)

PR #116, opened 2023-11-12
misc: fix spelling

jbampton:


jkesselm (approved):
Good catch. Yeah, we've got some old comments that really should be cleaned up...

PR #119, opened 2023-11-14
Maven build improvements

kriegaex:
* Use platform-independent path separators in `exec:exec`
* Reorganise .gitignore, add IntelliJ IDEA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,16 @@
-samples/target/ 
-serializer/target/ 
-xalan2jtaglet/target/ 
-xalan/target/
-target/
-build/
+# Maven
+target/ 
 **/dependency-reduced-pom.xml
+
+# Eclipse
+.settings/
+.project
+
+# IntelliJ IDEA
+.idea/
+
+# Xalan
+build/
 **/xsltc/compiler/XPathLexer.java
 **/xsltc/compiler/XPathParser.java
 **/xsltc/compiler/sym.java
-.settings/**
--- a/pom.xml
+++ b/pom.xml
@@ -16,6 +16,9 @@ command "mvn compile dependency:tree" so everything is in scope.
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <stylebook.classpath>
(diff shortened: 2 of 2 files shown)


kriegaex:
Like I said on the mailing list, I just wanted to make the Maven build run and then noticed that there was no more error creating assemblies. So for now, I did not dig any deeper. I wanted your feedback first, focusing on solving any possible problems you are still encountering first, before eventually we might look into more cosmetical things - the proviso always being that I even have sparetime to look into it.

jkesselm:
Understood. Thanks for even the quick glance! I have a bit more fine tuning to do, and I may hit something unexpected, but it seems to basically be under control for now.

One oddity I did encounter: When the binary-assembly module was set to run as type pom. It seemed to be overwriting its own pom.xml with one of the archive files. For now I've worked around that by setting the output type to jar even though it shouldn't be producing a jarfile we need, but at some point I will have to understand why that was happening.

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

kriegaex:
I suggest to make your parent POM just an aggregator with modules and not put any artifacts or artifact-creating plugins in there. It is better to create modules for things like docs generation or assembly building.

kriegaex:
> When the binary-assembly module was set to run as type pom. It seemed to be overwriting its own pom.xml with one of the archive files. For now I've worked around that by setting the output type to jar even though it shouldn't be producing a jarfile we need, but at some point I will have to understand why that was happening.

No, it does not overwrite the POM. I tried with your latest commit. Maybe in a previous version, you experimented with output directories or with special path and name for the dependency-reduced POM and accidentally ended up putting it into the base instead of the build directory.

jkesselm:
Thanks. Suspected that might be the case.

("Make it work. Make it good. Make it great." -- Steve Boies)

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

jkesselm:
Well... If I try to change distribution/pom.xml to `<packaging>pom</packaging>`, I *do* reproduce the problem of that file being overwritten. Revert that file, change it back to `<packaging>pom</packaging>`, and the problem goes away. 

The workaround of producing a near-empty jarfile is good enough for me, for now. But it would be good to understand whether the overwrite is a Maven/plugin bug, or is me having done something stupid.

PR #120, opened 2023-11-14
Reformat Maven POM and assembly XML, no functional changes

kriegaex:
- Get rid of mixed tabs and spaces, use 2 spaces for indentation uniformly instead.
- Use line width 120 with very few exceptions.
- Use a uniform way to format multi-line XML comments, putting opening and closing tag parts on separate lines, indenting the text content by 2 spaces.

This can subsequently be the baseline for further commits, both by @kubycsolutions and external PR contributors.

Sorry for maybe seeming nitpicky, but when scrolling through the POMs in my IDE, it seemed somewhat messy and hard to parse in my limited brain. It might be a personal preference only, but I do like uniform indentation. As the Java code mostly seems to use 2 spaces and the POMs also partly followed that pattern, I chode 2 spaces too - maybe also, because that is my personal preference.

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -1,7 +1,8 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" 
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
+
   <parent>
     <groupId>xalan</groupId>
     <artifactId>xalan-project</artifactId>
@@ -9,17 +10,18 @@
   </parent>
 
   <artifactId>distribution</artifactId>
-  
-<!-- PROBLEM: Something has been overwriting this pom. I don't know
-       whether making the packaging "pom" had anything to do with that,
-       but changing it may have helped. 
+
+  <!--
+    PROBLEM: Something has been overwriting this pom. I don't know whether making the packaging "pom" had anything to do
+    with that, but changing it may have helped. 
   -->
   <packaging>jar</packaging>
- 
+
   <name>distribution</name>
-  
-  <!-- NOTE: These dependency declarations are only required to sort
-       this project to the end of the queue in the multimodule build.
+
+  <!--
+    NOTE: These dependency declarations are only required to sort this project to the end of the queue in the
+    multimodule build.
   -->
   <dependencies>
     <dependency>
@@ -38,12 +40,12 @@
       <version>2.7.3</version>
     </dependency>
   </dependencies>
- 
+
   <build>
     <plugins>
       <plugin>
         <artifactId>maven-assembly-plugin</artifactId>
-	<version>3.6.0</version>
+        <version>3.6.0</version>
         <executions>
           <execution>
             <id>distro-assembly</id>
@@ -59,8 +61,6 @@
           </execution>
         </executions>
       </plugin>
-
-
 
     </plugins>
   </build>
--- a/distribution/src/assembly/bin.xml
+++ b/distribution/src/assembly/bin.xml
@@ -1,12 +1,13 @@
-<!-- Maven assembly plugin configuration for executable packaging 
-     See https://maven.apache.org/plugins/maven-assembly-plugin/faq.html#module-binaries
-     TODO: Should include xercesImpl and xml-apis jarfiles? 
-	Not regex, bcel, commons
-	And those should be moved to xalan-X_version/
+<!--
+  Maven assembly plugin configuration for executable packaging.
+  See https://maven.apache.org/plugins/maven-assembly-plugin/faq.html#module-binaries.
+  TODO: Should include xercesImpl and xml-apis jarfiles? 
+  Not regex, bcel, commons.
+  And those should be moved to xalan-X_version/.
 -->
 <assembly xmlns="http://maven.apache.org/ASSEMBLY/2.2.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.2.0 http://maven.apache.org/xsd/assembly-2.2.0.xsd">
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.2.0 http://maven.apache.org/xsd/assembly-2.2.0.xsd">
   <id>bin</id>
   <formats>
     <format>zip</format>
@@ -28,10 +29,9 @@
       </binaries>
     </moduleSet>
   </moduleSets>
-  <!-- Should also include source for all the samples, and
-       all the documentation, for emulation of
-       Ant behavior. Rename site to docs for back-compatibility.
-       Should pick up version from variable...?
+  <!--
+    Should also include source for all the samples, and all the documentation, for emulation of Ant behavior. Rename
+    site to docs for back-compatibility. Should pick up version from variable...?
   -->
   <fileSets>
     <fileSet>
@@ -43,8 +43,8 @@
       <directory>../samples</directory>
       <outputDirectory>${project.parent.name}_${project.parent.version}/samples</outputDirectory>
       <excludes>
-	<exclude>target/**</exclude>
-	<exclude>src/site/**</exclude>
+        <exclude>target/**</exclude>
+        <exclude>src/site/**</exclude>
       </excludes>
     </fileSet>
   </fileSets>
--- a/pom.xml
+++ b/pom.xml
@@ -1,18 +1,21 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <packaging>pom</packaging>
-  <!-- KNOWN ISSUES:
-       "Production" jar/tar files?
-       xalan-test integration
-
-NOTE: To get dependency tree from a multi-module project, use the 
-command "mvn compile dependency:tree" so everything is in scope.
-  -->
-  
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
+
+  <!--
+    KNOWN ISSUES:
+    "Production" jar/tar files?
+    xalan-test integration
+    
+    NOTE: To get dependency tree from a multi-module project, use the command "mvn compile dependency:tree" so
+    everything is in scope.
+  -->
+
   <groupId>xalan</groupId>
   <artifactId>xalan-project</artifactId>
   <version>2.7.3</version>
+  <packaging>pom</packaging>
 
   <properties>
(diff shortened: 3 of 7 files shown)


kubycsolutions:
I think my conclusion right now is: normalized indentation to spaces only and 2-column steps is accepted as a definite improvement.

Increasing line length by 50% as a general practice I'm reserving judgement on, especially when the entire Xalan codebase is styled to lines of about  80 characters. If folks want to write long lines, then as long as they're reasonably readable I won't object; if they incidentally restyle as part of a substantive change, ditto (as long as they're willing to extend me the same courtesy). But I'm not yet seeing a need to force a change before it happens organically.

You _do_ realize that the reason for reducing indentation stops from 8 spaces to 4 to less was to better fit 80-column terminals and narrower windows, right? Asking for both that and longer lines is a touch inconsistent historically. But de gustibus non disputandum est, and coding style beyond basic legibility is definitely a matter of taste. I can work with almost anything except the guy who insisted that in his C code semicolon went before the start of a statement rather than at its end.

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.
From: Alexander Kriegisch ***@***.***>
Sent: Monday, November 13, 2023 11:04:44 PM
To: apache/xalan-java ***@***.***>
Cc: Joseph Kesselman ***@***.***>; Mention ***@***.***>
Subject: [apache/xalan-java] Reformat Maven POMs, no functional changes (PR #120)


  *   Get rid of mixed tabs and spaces, use 2 spaces for indentation uniformly instead.
  *   Use line width 120 with very few exceptions.
  *   Use a uniform way to format multi-line XML comments, putting opening and closing tag parts on separate lines, indenting the text content by 2 spaces.

This can subsequently be the baseline for further commits, both by @kubycsolutions<https://github.com/kubycsolutions> and external PR contributors.

Sorry for maybe seeming nitpicky, but when scrolling through the POMs in my IDE, it seemed somewhat messy and hard to parse in my limited brain. It might be a personal preference only, but I do like uniform indentation. As the Java code mostly seems to use 2 spaces and the POMs also partly followed that pattern, I chode 2 spaces too - maybe also, because that is my personal preference.

You can view, comment on, or merge this pull request online at:

  https://github.com/apache/xalan-java/pull/120

Commit Summary

  *   816d616<https://github.com/apache/xalan-java/pull/120/commits/816d6163797fd0a7da3a099112c67bc1b6dc4ec6> Reformat Maven POMs, no functional changes

File Changes

(6 files<https://github.com/apache/xalan-java/pull/120/files>)

  *   M distribution/pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-ac19e9d7369629bf473b9b6c514012d5fecc319f98a6b9b57855f3202e970610> (30)
  *   M pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8> (402)
  *   M samples/pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-eb8baa2a7ef27be22bc6fa693b229395e6ae824a95ccde07202d8d97b91f7740> (89)
  *   M serializer/pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-2cb43b87715a28c31056f717bee375907a7d6b05db40dd14899346909d1662d5> (92)
  *   M xalan/pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-8fc15af17bf692a644c8fc77eb847323a47d0803f45aac0952255b991195fcbe> (150)
  *   M xalan2jtaglet/pom.xml<https://github.com/apache/xalan-java/pull/120/files#diff-de7f09ef626f3db137f8fb539236aab2a328faea4ad18322f3054b376cb66aa1> (103)

Patch Links:

  *   https://github.com/apache/xalan-java/pull/120.patch
  *   https://github.com/apache/xalan-java/pull/120.diff

kriegaex:
Sorry, I thought I saw longer lines in other places, too. BTW, at the age of 52 years I am very much aware of the historic roots of the 80 character limit. But I am also aware of the fact that next to no developer, not even old guys like us, work with 80 character terminals anymore. Even in vi or Nano, usually I work with way more characters per line, unless I need to work in a Linux root terminal. But there, I do not write programs but maybe edit some config file, if my Linux kernel refuses to boot. but even then, I can scroll horizontally. I think, us older semesters should not impose an 80 character limit on younger developers just because of the good old days. 120 characters is widely accepted in many OSS projects and still narrow enough fore me to fit more characters and a few tool windows on screen in my IDE. Besides, I was not reformatting the Java code, just a handful of POMs.

jkesselm:
I've applied most of your changes manually, since we have agreed that we disagree on a few details of formatting style and since some obsolete comments could be ripped out entirely rather than just reflowed/re-indented. If that satisfies your intent, we can close this PR.  If not, please merge and reconcile so it's focused on the remaining differences.

Thanks!

jkesselm:
BTW, I've confirmed the problem in distribution/pom.xml. If I change the `<package>` type to pom, it overwrites itself with some form of binary. If I leave it as jar, it runs fine and the file is unharmed. So I have a workaround, but no idea why it would misbehave.

kriegaex:
Thanks, the uniform indentation is much better now. If you like vertical scrolling so much, keep your 80 character limit. I can live with it.

kriegaex:
> BTW, I've confirmed the problem in distribution/pom.xml. If I change the `<package>` type to pom, it overwrites itself with some form of binary. If I leave it as jar, it runs fine and the file is unharmed. So I have a workaround, but no idea why it would misbehave.

I cannot reproduce it. You should get to the bottom of this irregular behaviour. I tried `mvn clean verify` on Maven 3.9.2. Please explain how to exactly reproduce this.

kriegaex:
I have a hunch: Remember when I said that you should not define plugins globally in the parent POM, at least not ones you only need in some modules? I see that you have quite a list of stuff going on there. For the source assemblies, you can get away with it, using the special `runOnlyAtExecutionRoot` parameter. But that kind of special parameter is e.g. unavailable for Maven Shade, which is also defined in the parent POM and therefore executes for each module, each tinme saying "Replacing original artifact with shaded artifact". the original artifact for `pom` artifact types probably is the _pom.xml_ file itself, which is why it might get overwritten in your case, even though not in mine. Assembly for the source archive also runs for each module, albeit with a "Skipping the assembly in this project because it's not the Execution Root" message.

kriegaex:
As far as I can see, the only module in which Shade makes sense is the `xalan` module anyway, and there you redundantly defined the plugin already. It also is in the `samples` and `serializer` modules, where IMO it adds no value. Can you please explain why you think you want to shade each artifact? Is it even necessary for the `xalan` module, if you build a binary assembly later anyway? Usually, Shade or Assembly are only used, if you want to build a stand-alone application that should include all of its dependencies, to be able to run it from the CLi with an easy Java command.

kriegaex:
I think, you should also build the source distro in the `distribution` module, just like the binary one.

kriegaex:
Oh, I see what you are doing, using a centrally provided Apache plugin dependency for Assembly Plugin to be used on top level. Well, I kind of understand that this is easier than to custom-build your own assembly desceriptor. OTOH, this leads to directories being included that do not belong into the source distro. For example, I made some local changes to your project's Shade Plugin config, copying the resulting source/binary assemblies into a __tmp_ subdirectory before and after the change to be able to diff them afterwards. The whole __tmp_ subdirectory with dozens of MB of binaries ended up in the source distro. I.e., you need to be quite careful what is in your local workspace when running the build.

kubycsolutions:
With all the emphasis on "use Maven defaults", I presumed that the official Apache assembly descriptors were sufficiently robust to be preferred unless we had special needs.

If that isn't the case I can certainly reintroduce my own descriptor for the source bundle.

Might be moving in that direction anyway, to get everything that was in the Ant distribution to be present in the Maven version (modulo deliberate changes).

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

kubycsolutions:
Was moving in that direction. The code is evolving, and while I want elegance I also want to actually get it into use sometime soon
 The ideal is the enemy if the good enough.

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

kubycsolutions:
Again, my understanding was that Maven preferred central definition and local use. Certainly I'd rather have the things which are common to all the modules factored upward when that makes sense, to keep things reliably in sync.

It's starting to feel like there are a lot of exceptions to stated Maven philosophies. I can live with that, I understand that tools get gruesome because they grew some... But I don't _think_ anything I'm trying to achieve is that unusual for Apache tools.

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

kubycsolutions:
The exact command is in the mvnbuild.sh/.bat. I'm AFK right now, but I believe it's "mvn clean package"... Which, if I remember correctly, is a stage or two past verify. Package is the layer where jarfiles are produced, and thus where my distribution bin archive is built since it includes the ready-to-run jarfiles.

--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

kriegaex:
@kubycsolutions, it would be better to actually comment on GitHub and quote the relevant pars here than to respond via e-mail. When reading inline, I do not really see what you are quoting. The inline quotes that are expandable here are always full quotes including e-mail signature etc.

kriegaex:
> The exact command is in the mvnbuild.sh/.bat.

OK, I always use `mvn` directly, either from the IDE or the console. No extra wrapper scripts.

> I believe it's "mvn clean package"... Which, if I remember correctly, is a stage or two past verify.

No, the `verify` phase is after `package` and also executes post-package integration tests, if any. Only later I noticed that this project is completely devoid of any automated tests, so I also would not have had to use `-DskipTests` for faster builds in the first place. 😉

Speaking of faster builds: I think, it would make sense to define Maven profiles for faster vs. more complete builds. If you actually work on the code in the future, you do not always want to build Javadoc and other types of documentation, distro JARs etc. Usually, you just want to compile and run automated tests, if any. Currently, the build is quite slow because of all the optional steps.

kriegaex:
Another question is if you really want to force every developer to run the build on JDK 8 just because of the taglet dependencies on old `com.sun` JDK packages, which have been superseded by others on JDK 9+. Basically, you have 3 options:
  * Continue as is.
  * Require JDK 9+ for the build and migrate the taglet thingy to the new API standard.
  * Add a JDK 9+ variant of the taglet code on top of the JDK 8 version and configure the build to use one on JDK 8 and the other on JDK 9+.

WDYT?

jkesselm:
Good catch!

Xalan currently promises to run on JRE 8. That doesn't absolutely require that we build on JDK 8, I suppose, but I want to retain the ability to do so.  For now, that rules out requiring 9+.

But we'll move forward someday. And users may want to build their own copies on newer versions.

It's orthogonal to the Maven migration, though. So I think the right thing to do is open it as a Jira Issue and get back to it.

Fourth option: make the taglet itself introspect and pick the right set of APIs.


--

() I still don't think HTML mail is a good idea
/\ but Outlook/Android is insisting. Need to
     change mail client.

jkesselm:
Currently, xalan-test is a separate package. The idea was that it might be used/adapted as a test suite for other XSLT processors, who might not want to download all of Xalan to get it. Integration/regression testing requires downloading both and invoking the test drivers. It was also used for xalan-c, which I believe has been moved to the Attic as unused and extremely outdated.

I'm actually leaning toward pulling some or all of the test framework back into the Xalan-J. But that's a different work item.

kriegaex:
> Xalan currently promises to run on JRE 8. That doesn't absolutely require that we build on JDK 8, I suppose, but I want to retain the ability to do so.

Then we need two variants of the Taglet classes, one using the old API and one the new one. The APIs need to be used in different ways, and the old classes are no longer available and therefore do not even compile on newer JDKs. The other way around, the new APIs are unavailable in JDK 8. I.e., duplicate taglet classes and conditional compilation using auto-activated, JDK-dependent profiles.

> For now, that rules out requiring 9+. But we'll move forward someday. And users may want to build their own copies on newer versions.

See above.

> It's orthogonal to the Maven migration, though.

That is true on the one hand, but OTOH maybe it is worth doing now. I have never used taglets at all, but it would be a nice puzzle to solve to adjust the taglet classes to the new API and to get the build running as described above. Next time I am stuck in another project and need an exercise to stretch my mind, I might look into it. I have actually begun to do so already, albeit superficially.

> So I think the right thing to do is open it as a Jira Issue and get back to it.

https://issues.apache.org/jira/browse/XALANJ-2709

> Fourth option: make the taglet itself introspect and pick the right set of APIs.

That is not a good option, because without reflection it would not work. See above.

jkesselm:
> That is not a good option, because without reflection it would not work. 

Who said without reflection? I'm not dogmatic about that; certainly it's a nit compared to BCEL.

But, yeah, switching at the Maven level is reasonable, arguably more so. I was just tossing reflection onto the table for completeness.

PR #121, opened 2023-11-14
misc: fix spelling

jbampton:


jkesselm:
The comment changes are fine.

My one concern about this is that you've fixed some field and method names. That could be a breaking change if customers have coded against the old names. 

I'd recommend looking at whether there's a way to proxy the new names with the old names, and deprecate those proxies. That gives users a development cycle's worth of warning that this change will occur.

(I don't think there are any instances of "occured" in the official Xalan, JAXP, or TrAX APIs... but if it isn't `private`, there *is* a risk someone is accessing it, whether they should or not.)

PR #125, opened 2023-11-18
[XALANJ-2710] Use resource files for version numbers

kriegaex:
See [XALANJ-2710](https://issues.apache.org/jira/browse/XALANJ-2710). Version numbers are now read from Maven-generated `pom.properties` files for
  - Xalan
  - Xalan Processor
  - Xalan Serializer

The 3 existing `*Version.java` classes have been adjusted correspondingly, the related `*.src` files deleted, because they are no longer necessary.

BTW, the properties files are found both in JARs and in the file system, if the Maven build has created them. Otherwise, a version number 0.0.0 will be the fallback.

~~@kubycsolutions, sorry for being lazy again. This PR was branched off of the one for XALANJ-2709, even though I could also have branched it off the Maven branch. But I hope that you can review and merge them sequentially in short order.~~ I have just rebased the PR on the Maven branch. You can review two PRs in any order now. After one of them will have been merged, I can rebase the other one on the new Maven branch head.

Please note one breaking change: Both Xalan and Xalan Processor used to return the same version string. Then why do we need two version classes in a single Maven module for them in the first place? Serializer did not mention the product name Xalan at all, so it could be any serializer. I changed it as follows:

Product | Old version string | New version string
--|---|--
Xalan | Xalan Java 2.7.3 | Xalan Java 2.7.3
Processor | Xalan Java 2.7.3 | Xalan Processor Java 2.7.3
Serializer | Serializer Java 2.7.3 | Xalan Serializer Java 2.7.3

---

Personal note: There are several things I do not like about the current state of affairs, but I did not want to change too much:

  1. There are 3 version classes. I think that, if designed correctly, a single such class in a shared module would suffice. Now, there is code and Javadoc duplication.
  2. The 3 existing classes used different ways to solve the same problem (constants vs. getter methods) and different javadoc styles. I harmonised the former (now all of them use getters), but did not touch the javadocs, because they are ugly anyway and need an overhaul. Not even the `@return` tag is used correctly or not at all. The formatting is weird and looks like a mess in HTML. It only caters to source code formatting, but even there I do not understand the weird indentation within the javadoc comments.
  3. The version classes could be quite small with just a helper method to read the properties file and find the version number there. But because the getters for parts of the version numberlike major/minor are public - why? does anyone use them? - it would have been a breaking change to delete them. To retain them, I had to parse the version number from Maven into its constituent parts using a regex, only to concatenate the parts later in the method used to retrieve the version number. This is all too convoluted for my taste.


jkesselm:
Great. I was thinking of moving in that general direction, as you know.

Code complexity: Historical accident. It got rattled off between other things, it worked well enough that it didn't get revisited. It's not exactly in anyone's critical path...

Multiple classes: Shipping in two jarfiles (serializer and processor), we wanted to allow for the possibility of out-of-synch versions, so we needed two classes with their own constants. Plus the merge of Sun's XSLTC compiler into IBM's Xalan-J code left us with some redundancies.  So at least two of the three classes make some slight sense -- though having three completely different implementations, I agree, is ugliness that should be cured.

I do NOT claim this code is as clean as it should be! It's been largely unattended for a very long time, and the process of pulling IBMers off it was also not clean.  (I was one of the last out...)

And yes, the javadoc is erratic at best; one of the many things on my own backlog is cleaning it up, since I probably remember the overall architecture of the beast better than most.

Can I suggest we open Issues to make sure we don't forget these points? I'm willing to do that for you, but I'm short on sleep right now and probably less than coherent so if I'm doing it I'd rather wait until tomorrow...

kriegaex:
I forgot to mention: when reviewing this, you will notice that I also added the capability to detect Maven `-SNAPSHOT` suffixes in version numbers.

Currently, in your Maven branch, there is a fixed release version 2.7.3 in all POMs. During development, I would rather have expected a snapshot version, which would then be changed to a final release version by the Maven Release Plugin or, if you choose do not use that, manually by Versions Maven Plugin. There would be a single commit for the final release, which then gets tagged correspondingly. Subsequently, the next snapshot version wouldbe set. I am suggesting such a standard process rather than having many commits with the same final release version in Git. Maybe the snapshots could also replace the development drops (something like 2.7.D5), unless you want to keep them and combine them with snapshot usage, which certainly is also a feasible option.

serializer/resources/org/apache/xml/serializer/Version.src
```diff
@@ -1,150 +0,0 @@
...
-
-/**
- * Administrative class to keep track of the version number of
- * the Serializer release.
- * <P>This class implements the upcoming standard of having
- * org.apache.project-name.Version.getVersion() be a standard way 
- * to get version information.</P> 
- * @xsl.usage general
- */
-public final class Version
```
garydgregory, line 31: Does this PR remove a public class? If so, that's a -1 for breaking binary compatibility.
Let's make sure the Maven build uses JApiCmp to check.

kriegaex: Would you mind reviewing instead of making premature assumptions? Thank you.

garydgregory: Hello @kriegaex 
My review is asking the question: Does this PR break binary compatibility? 

There is no sense in spending time looking at details unless we are standing on solid ground.

kriegaex: That is not a review, it is a cross-examination. If you feel that you are too busy to conduct a proper review, I suggest you better do something that has higher priority to you. I understand that nobody can be in several places simultaneously. Either way, I think I am too busy to answer this kind of question. You would already know the answer, if you had taken even a cursory glance at the list of changed files or read the PR description.

kriegaex:
FYI, I have rebased this PR on _xalan-java-mvn-refactored_ after XALAN-2709 was merged.

Improvement idea: I suggest that this project adopts a rule that instead of merge commits and ugly loops and looplets like this, ...

image

... we always require PRs to be rebased on the target branches they are to be merges into. Given that precondition, we can completely avoid merge commits for PRs and simply fast-forward each PR.

kriegaex:
@kubycsolutions, please note the new optimising commit 6b80f98c. The class could be even shorter, if it was indeed intentional to return "Xalan Java" instead of something more specific like "Xalan Processor Java" in class `XSLProcessorVersion`. Was my assumption correct or wrong that it makes sense to report different product names there? I just noticed the Javacod comment in `XSLProcessorVersion`, saying `@deprecated To be replaced by org.apache.xalan.Version.getVersion()`. This would rather imply that they report identical names on purpose. When that deprecated extra class will be removed is out of scope here, so I have kept it, of course. The only open question to me is: same or different product name?

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
+    try (InputStream fromJar = Version.class.getClassLoader().getResourceAsStream(POM_PROPERTIES_JAR)) {
+      if (fromJar != null) {
+        pomProperties.load(fromJar);
+        version = pomProperties.getProperty("version", NO_VERSION);
+      }
+      else {
+        try (FileInputStream fromFileSystem = new FileInputStream(POM_PROPERTIES_FILE_SYSTEM)) {
+          pomProperties.load(fromFileSystem);
+          version = pomProperties.getProperty("version", NO_VERSION);
+        }
```
vlsi, line 72: Can you clarify when this codepath could be triggered?
I cloned https://github.com/kriegaex/Maven_GenerateSourcesExample in IDEA 2023.2.3, and sure it adds `target/classes` when executing `Main`

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
+      }
+    }
+    catch (IOException e) {
+      new RuntimeException("Cannot read properties file to extract version number information: ", e)
+        .printStackTrace();
```
vlsi, line 77: The error message on the stderr would be hard to relate to Xalan.

I would suggest use `java.util.logging` for the logging instead of `System.err`. Then the users would be able to identify the source of the message and silence it without recompiling Xalan.

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
+      else {
+        developmentVersionNum = Integer.parseInt(matcher.group(4));
+      }
+      snapshot = matcher.group(5) != null && !matcher.group(5).isEmpty();
+    }
+    else {
+      System.err.println(
+        "Cannot match version \"" + version + "\" " +
+          "against expected pattern \"" + VERSION_NUMBER_PATTERN + "\""
+      );
```
vlsi, line 98: I suggest using `java.util.logging` instead of `System.err`.
The generic message `Cannot match version ...` would be hard to relate to Xalan, and it would be virtually impossible to silence.

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
    */
   public static String getVersion()
   {
      return getProduct()+" "+getImplementationLanguage()+" "
            +getMajorVersionNum()+"."+getReleaseVersionNum()+"."
-           +( (getDevelopmentVersionNum() > 0) ? 
-               ("D"+getDevelopmentVersionNum()) : (""+getMaintenanceVersionNum()));  
+           +( (getDevelopmentVersionNum() > 0) ?
+               ("D"+getDevelopmentVersionNum()) : (""+getMaintenanceVersionNum()))
+           +(isSnapshot() ? "-SNAPSHOT" :"");
```
vlsi, line 115: Should it return the original version from the property file rather than reconstruct it?
It looks like `return version` should yield the same result

serializer/src/main/java/org/apache/xml/serializer/Version.java
```diff
@@ -20,33 +20,97 @@
...
+  private static final String VERSION_NUMBER_PATTERN = "^(\\d+)[.](\\d+)[.](D)?(\\d+)(-SNAPSHOT)?$";
+  private static final String NO_VERSION = "0.0.0";
+
+  private static String version = NO_VERSION;
+  private static int majorVersionNum;
+  private static int releaseVersionNum;
+  private static int maintenanceVersionNum;
+  private static int developmentVersionNum;
+
+  private static boolean snapshot;
```
vlsi, line 52: It looks like `serializer/Version` logic duplicates `xalan/Version` can it be shared?
`xalan` depends on `serializer` anyway, so it looks like `xalan` could just reuse the logic from `serializer` instead of duplicating it.

vlsi (changes requested):

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
  * org.apache.xalan.processor.Version class.</P>
- * <P>See also: org/apache/xalan/res/XSLTInfo.properties for 
+ * <P>See also: org/apache/xalan/res/XSLTInfo.properties for
  * information about the version of the XSLT spec we support.</P>
  * @xsl.usage general
  */
 public class Version
 {
+  private static final String POM_PROPERTIES_JAR = "org/apache/xalan/version.properties";
+  private static final String POM_PROPERTIES_FILE_SYSTEM = "xalan/target/classes/" + POM_PROPERTIES_JAR;
```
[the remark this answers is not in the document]
kriegaex: Good catch! I added that, because I had tried something from the command line, probably without setting the proper classpath. When running from the IDE or from a hypothetical unit or integration test, the corresponding code branch should never be entered. I am going to remove it.

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -20,36 +20,99 @@
...
+  private static final String VERSION_NUMBER_PATTERN = "^(\\d+)[.](\\d+)[.](D)?(\\d+)(-SNAPSHOT)?$";
+  private static final String NO_VERSION = "0.0.0";
+
+  private static String version = NO_VERSION;
+  private static int majorVersionNum;
+  private static int releaseVersionNum;
+  private static int maintenanceVersionNum;
+  private static int developmentVersionNum;
```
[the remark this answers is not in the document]
kriegaex: Please note, that the source code for these version classes was taken directly from the old Ant `*.src` templates. I wanted to refactor with minimal changes, hence I did not touch these here. The same applies to the ugly javadocs and the existing, IMO superfluous public methods for each part of the version number. If someone wants to clean up these classes in a later stage, I sympathise.

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: See above, I will remove that stuff.

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: Xalan does not seem to use any logging frameworks, especially not `java.util.logging`. There is `commons-logging` on the classpath, but also unused in the code base, as far as I can see. I doe see lots of `System.out` and `System.err` references in the code. So I adjusted to that style. A utility class for version information is hardly a good place to introduce a logging framework into a product.

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: See above.

kriegaex: See above.

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: Did you read the PR description? Search for the paragraph starting with "The version classes could be quite small". Besides, for backward compatibility something like "D04" would be parsed to integer 4 and then returned as "D4" the next time. so far, so strange. But I want to be backwards compatible. Another developer already voiced concerns about backward compatibility of public API before.

serializer/src/main/java/org/apache/xml/serializer/Version.java
kriegaex: No, it does not. I had checked that before.

```text
--- dependency:2.8:tree (default-cli) @ serializer ---        
xalan:serializer:jar:2.7.3                                    
+- org.apache.bcel:bcel:jar:6.7.0:compile                     
|  \- org.apache.commons:commons-lang3:jar:3.12.0:compile     
+- commons-logging:commons-logging:jar:1.2:provided           
+- javax:javaee-api:jar:6.0:provided                          
+- javax.servlet:servlet-api:jar:2.5:provided                 
+- xerces:xercesImpl:jar:2.12.2:provided                      
\- xml-apis:xml-apis:jar:1.4.01:provided                      
```

You see that I already optimised `XSLProcessorVersion` to avoid duplicate code. Again, I am pointing you to my PR description. Search for the paragraph starting with "There are 3 version classes". All those concerns you seem to have were already considered by me. Please do read descriptions too, not just review code in isolation.

xalan/src/main/java/org/apache/xalan/Version.java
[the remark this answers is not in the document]
kriegaex: OTOH, the fields are final, so I might as well do it now.

kriegaex:
@vlsi, please note commit 9ab82a06

xalan/src/main/java/org/apache/xalan/Version.java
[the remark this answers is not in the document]
vlsi: Please note that old template classes did not have fields at all, and the fields appear only in the current pr.

Can you please check again?
My comrent was related to fields only, and I do not understand why you bring methods and javadocs.

xalan/src/main/java/org/apache/xalan/Version.java
vlsi: 04 -> 4 is a fair point

xalan/src/main/java/org/apache/xalan/Version.java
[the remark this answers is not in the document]
kriegaex: You are right, I misremembered. I added the fields, because I only wanted to parse the version once after reading the resource file. The getters then refer to them. I am going to make them final. Sorry for my confusion and thanks for asking again. Sorry, I am foing too many things in parallel, because these PRs stay open too long. Xalan is about 5% of what I do every day, and I only started last week. I hate context switches. This is of course only an explanation, not an excuse.

I brought up methods and javadocs, because I thought they are the same category as fields. I thought the fields existed before, remembering wrong.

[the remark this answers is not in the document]
kriegaex: Um, now I remember more about a few days back when I introduced the fields: I tried to make them final before. But this is not an option, because they are static, but they get their values later in one of the helper methods after parsing the resource file. Therefore, they have to be non-final. I cannot just initialise them later in a constructor like non-static final fields.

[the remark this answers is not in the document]
kriegaex: The only way around that would be to inline the static helper method `parseVersionNumber()` into the static initialiser block, but IMO that would be ugly and not worth the trade-off of making a few private fields static.

serializer/src/main/java/org/apache/xml/serializer/Version.java
kriegaex: @kubycsolutions, @vlsi, do you think these version classes justify creating an extra module `xalan-commons` or so at this point, just to get rid of duplicate code for those non-essential classes? If you both say "yes", I shall of course follow your advice and add one such module. But it seems to be out of scope of this PR, which is meant to migrate an existing Ant solution to a Maven one with a different approach. OTOH, it is never too early to apply the DRY principle. Up to you, but you have to agree for me to change it.

xalan/src/main/java/org/apache/xalan/Version.java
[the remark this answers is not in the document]
vlsi: > The only way around that would be to inline the static helper method parseVersionNumber() into the static initialiser block

The method is trivial, so having it inlined does not make the code uglier or better. It would be pretty much the same.

Have you considered extracting `Version(...)` record-like class with instance `final` fields?

* It would allow `final` fields with a plain Java code without static initializer blocks and so on
* It would make the code easier to test. Well, why not add a unit test or two for the regexp in the parsing logic?
* It would enable reusing the parse logic across "version" classes
* It might even make the code easier to read

serializer/src/main/java/org/apache/xml/serializer/Version.java
vlsi: @kriegaex , sure xalan had no third-party dependencies before, so all these commons-logging and bcel is something unexpected. I assume they were introduced in the maven migration by mistake.
Here are the proper dependencies: https://repo1.maven.org/maven2/xalan/xalan/2.7.1/xalan-2.7.1.pom

vlsi: Frankly, I doubt that "version class" alone justifies introducing a `xalan-commons` module. `xalan:serializer` has always been a mandatory dependency, so placing `Version` class into `serializer` would be less invasive.

jkesselm: Unfortunately, separate Version classes are necessary for the processor and Serializer. These are separate jarfiles because they are sometimes used independently. Hence we have to be able to ask each for it's version when the other is not available, which makes factoring out a common superclass across them undoable unless it's in a third jarfile. There's also the possibility of an out-of-sync pair, which we want to make visible. I can't see introducing a jarfile just for this, and I don't know of a Version in our dependencies which we could subclass productively. 

But if there's a separate Version for the xsltc compiler (I'm away from my desk and can't check easily, but I think there was/is),  _that_ can just proxy or subclass Xalan's Version, since they're the same jarfile and now report the same release version. That separation would exist only for historical reasons; the compiler started as a separate project before we adapted it to share Xalan's front end. I'd vote for deprecating this in favor of the Xalan Version, if it does exist. 

Of course the fine-tuning can be done as a separate PR, if folks are amenable to that. I've just been holding off merge because I haven't found cycles to review it myself (I'll try to do that today) and because there's still some productive discussion here. Alex, if you'd prefer further noncritical changes to be a separate PR,  we can open a Jira item for that and freeze this PR except for actual problems within its intended scope; I don't want to keep you hanging on this longer than necessary.

vlsi: @jkesselm , xalan.jar can't work without serializer.jar as it will almost definitely produce `NoClassDefFoundError`
Here is the relevant issue https://issues.apache.org/jira/browse/XALANJ-2649

xalan/src/main/java/org/apache/xalan/Version.java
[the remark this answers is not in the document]
kriegaex: I thought about it and decided not to inline it. Thanks for your input. I am also at this stage not going to write any tests, simply because this project does not contain any tests and it is way beyond the scope of this PR to introduce them. Remember, this is a PR to be merges into another PR which has the main purpose to convert the Ant build to Maven. I do not want to burden that PR by even more unrelated changes. The current test suite lives in a separate project, in which we would have to create shadow PRs in shadow branches to sync the tests to the branches here. Moreover, currently the classes I am changing here are untested in _xalan-test_.

Don't get me wrong, I am all in for test automation. I even teach classes about it. But like I said elsewhere, integrating tests here is scheduled to come after the Maven cutover. It has not even been discussed if the tests are to be migrated like they are implemented now, with the home-brew test framework they seem to use, or migrated to something like JUnit 5 Jupiter or Spock.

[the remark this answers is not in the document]
kriegaex: BTW, inlining the method would also make it next to untestable in the future. Testing code in a static block needs advanced tools, unless you are content with just testing the side effects. Of course, an even better solution would be a singleton pattern instead of static methods, then testability would be even better. But I really just want to help Joe to get his version classes from Ant back in the Maven build and not over-engineer the most unimportant utility class to be an order of magnitude cleaner than the rest of the code in this project. I really think that at the moment we have more important issues to deal with. You have my proof of concept, which works just fine. Merge it or do not merge it, up to you guys.

jkesselm:
Quick test: Compiling with 21, running as 1.8

### The good part:

Looks ok when invoked from command line with java -cp <path>/Version. 
- MINOR quibble: It might be nice to have this exception explain that the most likely cause is that the version number is available only when a Maven build is done. We can add that after the merge; it's not a merge blocker.

### Less good:

Ran Maven from commandline, confirmed the pom.properties exists in the expected places. Refreshed xalan-project in Eclipse; pom.properties is seen. Pointed at and ran org.apache.xml.serializer.Version. _It still reports the exception and 0.0.0._  Oops?

This appears to be because after importing the Maven project, the default working directory for directly executing this class is its own module, ${workspace_loc:serializer}. not the parent module..., so the path starting with serializer/ is not resolving.  I suspect a similar issue would exist if the non-jar copy of our Version was invoked from another project in Eclipse; current directory there shouldn't be the xalan-project root either... so just adding another fallback path without serializer/ doesn't look like a good solution.

I don't know how other IDEs handle current directory, but I wouldn't be surprised if the same issue exists there. Counting on paths relative to current directory just isn't stable. If we could do a path relative to the class's location, maybe, but...

- We could just say "Version doesn't work reliably in the IDE unless you're running against the jarfiles". I think folks might consider that a regression.
- We _could_ go back to filtering the Version .java from templates. Ugly but it eliminates the context sensitivity.
- I don't have a third answer offhand, and it's past bedtime. Lemme sleep on it and see if anything materializes.

Oh well. "If it was easy, they wouldn't need people like us."

jkesselm:
(digression re userIDs: For Apache stuff, I should be using jkesselm... but I don't always remember to log in that way on Git when I've been doing other things. Couldn't convince Apache to change my ID there to keshlam; they want to keep me associated with my old work and I can't argue with that. Long story, but I answer to either Joe or Keshlam, and the nickname is surprisingly close to unique on teh interwebz.)

kriegaex:
I did the unthinkable: open Eclipse. 😉

Then I deleted all `.project`, `.classpath` and `.settings` files and folders I found in my workspace - it looks like somehow you committed some of those to Git, or maybe they are old leftovers from Ant times. I think, project configs should be kept out of SCMs, which is why I added both Eclipse and IDEA folders and files to `.gitignore`.

Anyway, next I imported the already cloned repository as a **Maven project** like this:

image

Automatic Eclipse builds work fine, and correct version numbers are reported. Can you please try like this? If it also works for you, I will delete the Eclipse project files from Git and push an update to this PR.

@jkesselm, well, I found your other user ID by browsing the commit history here. You commit under that other name a lot.

jkesselm:
UserIDs: Ah. Probably left the git CLI logged in as Keshlam. (Kubyc's on its way toward becoming my primary home domain, I think.)

.whatever files: Old habits. Back when we were writing Xalan, checking in IDE configurations was common practice as a way of making sure everyone was using the same formatting rules and so on. I think I've gotten to the point where I'm willing to concede that as no longer practical; it took me a while to adjust to that.

Deleted xalan projects from Eclipse. Exited Eclipse. 
mvn clean, to ensure against cross-contamination.
Deleted .settings, .classpath, .project in xalan-project and its subdirs.
Restarted Eclipse. (It's currently using OpenJDK 21, though compiling in 1.8 mode).
Import Maven xalan-project. Open serializer Version class. Run that class as Java App (without changing any run parameters).
```
java.lang.RuntimeException: Cannot read properties file to extract version number information: 
	at org.apache.xml.serializer.Version.readProperties(Version.java:74)
	at org.apache.xml.serializer.Version.<clinit>(Version.java:55)
Caused by: java.io.FileNotFoundException: serializer/target/maven-archiver/pom.properties (No such file or directory)
```
Project/Update: No change.
Project/Clean: No change.

Looking at filesystem, I do see that we created
    serializer/target/classes/META-INF/maven/xalan/serializer/pom.properties
    serializer/target/maven-archiver/pom.properties

Insert debugging in readProperties to print java.io.File("").getAbsolutePath()
    Current directory when launched this way defaults to /usr/keshlam/git/xalan-project/serializer
    So file being fetched is /usr/keshlam/git/xalan-project/serializer/serializer/...etc...
If I explicitly change cd to xalan-project,  it tries to fetch /usr/keshlam/git/xalan-project/serializer/...etc...
   Closer, but still looking in source trees when you generated pom.properties into target.

So it looks like, when serializer Version is invoked directly from Eclipse, its default current directory is the root of the source tree of serializer, rather than from object tree root of xalan-package as your path assumes. We could generate pom.properties into source and have build copy it to object...

... butI'm still nervous about what happens when some other project sets its own current directory before checking values from org.apache.xml.serializer.Version. I don't think merely making sure it's in both source and object trees fixes that one; the relative path will be wrong. Unless we can somehow create a path relative to the class's file, which is starting to get signifcantly ugly.

There must be some way to get path relative to the current class, else fetching property files wouldn't work as expected, right? Or are folks always putting property files in known locations so they don't have to be relative, and using the library functions to load them from jarfiles so those can avoid the current-directory issue?

If it happens, it must be possible...?

jkesselm:
It's certainly possible that my Eclipse setup is weird. I don't think so, but... And if my analysis is close to correct, it might not be just an Eclipse issue.

I should see what happens if I explicitly tell Eclipse to run this with current directory in target/, just as crosscheck.

kriegaex:
@jkesselm, we need a live debugging session with desktop sharing. I have a Teamviewer licence. This way, we expend too much time.

The current directory should be irrelevant, if the classpath is correct. The version classes load the resource files as classpath resources. Are you testing on the current version of this PR, 613dc3380c948e177bce161eb3bc01392c005b58?

kriegaex:
Update: I just removed the Eclipse files in c89049ca915116c75426d6135288cef2347747d7, FWIW.

kubycsolutions:
I'm a bit fragged at the moment, but I'll look again. I *should* be current.

--

kriegaex:
Look, I would like to get all open PRs off the table ASAP, because otherwise it gets a bit difficult to start anything new. Written conversation has low bandwidth. Again, I am offering online chat, video calls, desktop sharing. My contact data (Telegram, WhatsApp, Gitter) are on my [GitHub profile](https://stackoverflow.com/users/1082681/kriegaex?tab=profile) (expand text and scroll to the bottom).

I cannot just commit on your Maven branch, because I am not a Xalan committer. But like this, we are moving very slowly, and actually I had not intended to stick around for so long on my temporary stint here. But we have started collaborating already, so I want to contribute a little bit more before I focus on other projects again. It would be nice to see the Maven cutover finished in your project.

kriegaex:
For the record: It was an issue with Git branches. I am glad we were able sort it out together in our little desktop sharing session. So good to know you personally now, @jkesselm. 😊

jkesselm:
Absolutely agreed; I've already taken more of your time than I deserve.

Outcome of telecon: Oops. I was running an outdated version of the code; I must have fat-fingered a git command.

Once the right code was in place, things worked as intended. I want to do a few quick tests as proper review, but I expect to merge very soon. Maybe not tomorrow given family commitments for the holiday, but by Friday unless something very unexpected happens.

kriegaex:
@jkesselm, please note: I have just squashed two pairs of commits in order to straighten out the commit history a bit before you merge. I have seen that you work from the command line when using Git, not from an IDE like I do. Therefore a hint: You want to issue a simple `git rebase` command in order to refresh your local branch. As you did not commit anything on top of my old changes locally, there is nothing to rebase interactively, it should just work seamlessly. The code has not changed, there are just two fewer commits than before.

kubycsolutions:
Thanks for the heads-up.

--

PR #129, opened 2023-11-25
Make version classes work when Xalan is on bootstrap classpath

kriegaex:
Sometimes, users put Xalan is on the bootstrap class loader (BCL). In this case, `Version.class.getClassLoader()` returns null. I.e., we should fall back to the system class loader, because there is no `Classloader` instance for the BCL (native code). Due to class loader hierarchy, however, the resource will also be found when asking for it from a level below the BCL.

**Update:** `Version.class.getResourceAsStream("version.properties")` does what we need, because it uses a relative resource path. The rest is doe internally by the JDK.

Relates to #125 and XALANJ-2710. Fixes corresponding smoke tests on xalan-test.

@jkesselm

### DIFF
--- a/serializer/src/main/java/org/apache/xml/serializer/Version.java
+++ b/serializer/src/main/java/org/apache/xml/serializer/Version.java
@@ -55,7 +55,16 @@ public final class Version
 
   private static void readProperties() {
     Properties pomProperties = new Properties();
-    try (InputStream fromResource = Version.class.getClassLoader().getResourceAsStream(POM_PROPERTIES_PATH)) {
+    ClassLoader classLoader = Version.class.getClassLoader();
+    if (classLoader == null) {
+      // Oops! Someone put Xalan is on the bootstrap class loader (BCL) -> fall
+      // back to the system class loader, because there is no Classloader
+      // instance for the BCL (native code). Due to class loader hierarchy,
+      // however, the resource will also be found when asking for it from a
+      // level below the BCL.
+      classLoader = ClassLoader.getSystemClassLoader();
+    }
+    try (InputStream fromResource = classLoader.getResourceAsStream(POM_PROPERTIES_PATH)) {
       if (fromResource != null) {
         pomProperties.load(fromResource);
         version = pomProperties.getProperty("version", NO_VERSION);
--- a/serializer/src/main/java/org/apache/xml/serializer/Version.java
+++ b/serializer/src/main/java/org/apache/xml/serializer/Version.java
@@ -36,7 +36,6 @@ import java.util.regex.Pattern;
  */
 public final class Version
 {
-  private static final String POM_PROPERTIES_PATH = "org/apache/xml/serializer/version.properties";
   private static final String VERSION_NUMBER_PATTERN = "^(\\d+)[.](\\d+)[.](D)?(\\d+)(-SNAPSHOT)?$";
   private static final String NO_VERSION = "0.0.0";
 
@@ -55,16 +54,17 @@ public final class Version
 
   private static void readProperties() {
     Properties pomProperties = new Properties();
-    ClassLoader classLoader = Version.class.getClassLoader();
-    if (classLoader == null) {
-      // Oops! Someone put Xalan is on the bootstrap class loader (BCL) -> fall
-      // back to the system class loader, because there is no Classloader
-      // instance for the BCL (native code). Due to class loader hierarchy,
-      // however, the resource will also be found when asking for it from a
-      // level below the BCL.
-      classLoader = ClassLoader.getSystemClassLoader();
-    }
-    try (InputStream fromResource = classLoader.getResourceAsStream(POM_PROPERTIES_PATH)) {
+    // IMPLEMENTATION NOTE: Class.getResourceAsStream uses a *relative* path by
+    // default, in contrast to Classloader.getResourceAsStream, which uses an
+    // *absolute* one. This is not clearly documented in the JDK, only
+    // noticeable by the absence of the word "absolute" in
+    // Class.getResourceAsStream javadocs. For more details, see
+    // https://www.baeldung.com/java-class-vs-classloader-getresource.
+    //
+    // Because we expect the properties file to be in the same directory/package
+    // as this class, the relative path comes in handy and as a bonus is also
+    // relocation-friendly (think Maven Shade).
+    try (InputStream fromResource = Version.class.getResourceAsStream("version.properties")) {
       if (fromResource != null) {
         pomProperties.load(fromResource);
         version = pomProperties.getProperty("version", NO_VERSION);
--- a/xalan/src/main/java/org/apache/xalan/Version.java
+++ b/xalan/src/main/java/org/apache/xalan/Version.java
@@ -58,7 +58,16 @@ public class Version
 
   private static void readProperties() {
     Properties pomProperties = new Properties();
-    try (InputStream fromResource = Version.class.getClassLoader().getResourceAsStream(POM_PROPERTIES_PATH)) {
+    ClassLoader classLoader = Version.class.getClassLoader();
+    if (classLoader == null) {
+      // Oops! Someone put Xalan is on the bootstrap class loader (BCL) -> fall
+      // back to the system class loader, because there is no Classloader
+      // instance for the BCL (native code). Due to class loader hierarchy,
+      // however, the resource will also be found when asking for it from a
+      // level below the BCL.
+      classLoader = ClassLoader.getSystemClassLoader();
+    }
+    try (InputStream fromResource = classLoader.getResourceAsStream(POM_PROPERTIES_PATH)) {
       if (fromResource != null) {
         pomProperties.load(fromResource);
           version = pomProperties.getProperty("version", NO_VERSION);
--- a/xalan/src/main/java/org/apache/xalan/Version.java
+++ b/xalan/src/main/java/org/apache/xalan/Version.java
@@ -39,7 +39,6 @@ import java.util.regex.Pattern;
  */
 public class Version
 {
-  private static final String POM_PROPERTIES_PATH = "org/apache/xalan/version.properties";
   private static final String VERSION_NUMBER_PATTERN = "^(\\d+)[.](\\d+)[.](D)?(\\d+)(-SNAPSHOT)?$";
   private static final String NO_VERSION = "0.0.0";
 
@@ -58,16 +57,17 @@ public class Version
 
   private static void readProperties() {
     Properties pomProperties = new Properties();
-    ClassLoader classLoader = Version.class.getClassLoader();
-    if (classLoader == null) {
-      // Oops! Someone put Xalan is on the bootstrap class loader (BCL) -> fall
-      // back to the system class loader, because there is no Classloader
-      // instance for the BCL (native code). Due to class loader hierarchy,
-      // however, the resource will also be found when asking for it from a
-      // level below the BCL.
-      classLoader = ClassLoader.getSystemClassLoader();
-    }
-    try (InputStream fromResource = classLoader.getResourceAsStream(POM_PROPERTIES_PATH)) {
+    // IMPLEMENTATION NOTE: Class.getResourceAsStream uses a *relative* path by
+    // default, in contrast to Classloader.getResourceAsStream, which uses an
+    // *absolute* one. This is not clearly documented in the JDK, only
+    // noticeable by the absence of the word "absolute" in
+    // Class.getResourceAsStream javadocs. For more details, see
+    // https://www.baeldung.com/java-class-vs-classloader-getresource.
+    //
+    // Because we expect the properties file to be in the same directory/package
+    // as this class, the relative path comes in handy and as a bonus is also
+    // relocation-friendly (think Maven Shade).
+    try (InputStream fromResource = Version.class.getResourceAsStream("version.properties")) {
       if (fromResource != null) {
         pomProperties.load(fromResource);
           version = pomProperties.getProperty("version", NO_VERSION);


serializer/src/main/java/org/apache/xml/serializer/Version.java
```diff
@@ -55,7 +55,16 @@ public final class Version
...
+    ClassLoader classLoader = Version.class.getClassLoader();
+    if (classLoader == null) {
+      // Oops! Someone put Xalan is on the bootstrap class loader (BCL) -> fall
+      // back to the system class loader, because there is no Classloader
+      // instance for the BCL (native code). Due to class loader hierarchy,
+      // however, the resource will also be found when asking for it from a
+      // level below the BCL.
+      classLoader = ClassLoader.getSystemClassLoader();
+    }
+    try (InputStream fromResource = classLoader.getResourceAsStream(POM_PROPERTIES_PATH)) {
```
vlsi, line 67: This is [all wrong](https://www.youtube.com/watch?v=ie2V8fWqer4)

Have you considered ~using a template Version.java~  `Version.class.getResourceAsStream(...)`?

jkesselm: Vlsi@: templating has been considered. It is currently considered a fallback plan.

jkesselm: I take it you didn't test; trying it with `Version.class.getResourceAsStream()` doesn't return a usable version string even when Version is invoked directly with `java -cp build/xalan.jar org.apache.xalan.Version`

jkesselm:
My test environment seems to be tangled somehow. I may be a while straightening it out. Apologies in advance for any delayed responses.

jkesselm:
As a sanity check, just to make sure there weren't other complications getting in the way, I tried reverting the $Version.java classes to the templated/hardcoded implementations used in the ant build.

With those (and with lib/ copied back into the xalan-java directory, to satisfy xalan-test's dependencies -- see the other PR re reconstructing these until xalan-test can manage them on its own), smoketest does run cleanly.

Now I'm in a position again to do a clean test of this PR. 

If it works, I'll merge. But we do want to remember that the templated approach is there as a fallback if this turns out to have other subtle interactions.

(And, yes, I'll open a work item ontaking the bootclasspath stuff out of smoketest. But since our customers were used to having to do that dance, I think we do need to be able to tolerate running in that mode, annoying as it is, at least for one deprecation cycle. We can have the fallback write a reminder out to stderr, or something of that sort, to help folks realize they should fix it.)

serializer/src/main/java/org/apache/xml/serializer/Version.java
kriegaex: > This is all wrong

Another judgemental and and utterly unhelpful comment. Snap out of it, please, dear @vlsi. If something is wrong in your opinion, please explain why and show how it is done right.

> Have you considered `Version.class.getResourceAsStream(...)`?

Not just considered, but also tried before I even switched to the `getClassLoader()` version. It failed, just like Joe said. Anyway, at least one thing you achieved with your unhelpful comment: I got curious why it had not worked, why the heck both `Class` and `Classloader` have the same method and what the difference between them is. Javadoc is pretty much useless, not explaining the difference in any meaningful way. But [Baeldung](https://www.baeldung.com/java-class-vs-classloader-getresource) does: `Class.getResourceAsStream` uses a *relative* path by default, in contrast to `Classloader.getResourceAsStream`, which uses an *absolute* one. This information enabled me to follow your "advice" and use `Version.class.getResourceAsStream("version.properties")`. It does exactly the same as my original solution, but does not require a null check and second try and even enabled me to inline the file name. As a bonus, it is also still relocation-friendly.

@jkesselm, I tested the new commit with normal and boot classpath and also with Maven Shade relocation. LGTM.

kriegaex:
@jkesselm,when merging, I suggest not to squash the commits, because that way we have two working alternatives in the Git history, both with explanatory commit and code comments.

vlsi:
What’s the point of leaving “system classloader fallback” commit?
I suggest removing it as the fallback was arbitrary, and Class.getResource is anyway much better.

serializer/src/main/java/org/apache/xml/serializer/Version.java
vlsi: @kriegaex , if I ask “use this exact code” you will definitely blame for me asking you to do something in “exactly the way I like”. Sure there might be unexpected issues. Of course Class.getResource.. works, but who knows what you tested else since you added exactly zero tests, and there was exactly zero CI executions. That is why I asked if you considered class-level getResourceAsStream.
If you failed to get it work, then just say so. It is frustrating when you blame me for both "asking to follow an exact route" and "asking if you considered a given workable alternative".

jkesselm:
> squashed

You were involved in the discussion leading to the Dev list exchange "Squash considered harmful?", in which we concluded that squash should be used only to discard true dead ends.

Request to squash declined as nitpicking.

kriegaex:
> What’s the point of leaving “system classloader fallback” commit?

I already explained:

>> because that way we will have two working alternatives in the Git history, both with explanatory commit and code comments

Moreover, we have documented the alternatives we have considered. Even better than that, we have a documented alternative, just in case in the future one of the people involved remembers this PR in a situation where we need to start from a class loader rather than from a know class. In that case, we would have to use my first strategy.

> Class.getResource is anyway much better

It is in this case, hence the current implementation.

serializer/src/main/java/org/apache/xml/serializer/Version.java
kriegaex: Good for you to know how I would "definitely" react to a meaningful code review.

Stop obsessing about tests in a project which does not have any tests. I explained in excruciating detail already that I am all in for tests, if they are part of the project, and why that is a separate (huge) PR. I also explained that I will not create shadow branches in the xalan-test repository. We would need one for the Maven branch and one for each of my PRs which currently are merged into Joe's Maven branch, not into the main branch, because the Maven cutover has not been done yet. How would anyone be able to track that sort of thing across projects? It is not my fault that Xalan-J does not treat tests as first-class citizens in code land. Instead of bickering about the symptom, why not lend a hand and migrate the test project into this one and thus eliminate the root cause of the problem?

One last time: Either be nice, or do not expect me to be. You want my respect? Treat me respectfully, or simply stop dropping your offensive comments into my PRs. Otherwise, Joe can thank you for me simply hopping off the project again. I am on the verge of doing so already, because the part of my spare time I dedicate to volunteer work in OSS development is too precious for having to spend most of it in debates with someone who obviously simply enjoys to argue.

vlsi: @kriegaex , let me be very explicit:
1. Please forget about xalan-test for now.
2. Please add junit-jupiter test scope dependency
3. Please create VersionTest.java file and add several test methods there. The methods should call public methods and verify they return a string that starts with "2."
4. Please factor regexp that parses version into package-private metod (or to a separate ParsedVersion class)
5. Please add ParsedVersionTese.java with a parameterized test that verifies several inputs. It could cover "d" versions or explicitly mention "d" versions will never be used (sure nobody should release those strange "d" versions now)

The tests I mean are like a couple of files 20 lines each.
Please behave. Your previous message sounds as if I asked you something complicated, however adding unit tests is trivial. If you see issues with adding unit tests, please tell me the reason explicitly. Sure it might be challenging to test "xalan on the boot classloader" with a regular junit jupiter test, however, a trivial test would still guard from regressions on the regular classloaders.

The lack of tests should not be your excuse for not adding tests for the code you change or create.

> why not lend a hand and migrate the test project into this one and thus eliminate the root cause of the problem?

0) Migrating xalan-test into xalan is not needed for adding unit test. At the same time, it would be a bad idea to test the regexp, property parsing, and basic Version features with integration tests
1) My PRs are either ignored or mentioned as "not a priority", so I am not going to create more until the situation improves
2) I am not interested in spending my time on Maven. Sure I can rework xalan-test with Gradle. However, I say it here **only because you ask**. I estimate the idea would not be accepted, so I don't even mention it on the dev list. At the same time, please note I offer a lot yet the team is just not willing to accept it.

kriegaex: First of all, @vlsi, I am glad to take suggestions from you, but no orders. Earlier we talked about test automation several times, and nowhere did anyone tell me to add test plugins or dependencies before the Maven cutover. @jkesselm added a JUnit 4 dependency for future use, but told me it was out of scope for the cutover. Now you are sort of commanding me to do the opposite, i.e. use JUnit 5 (Jupiter) instead of 4 and add tests instead of waiting for the cutover. Whom should I follow now? Are you in any position to speak authoritatively and command me to do such things? You are not even working actively on the Maven branch, because in your opinion Maven is a (quote) "waste of time".

Theoretically assuming for a minute, I would add tests separately from xalan-test, what would be the consequence? Tests would diverge, because they would be maintained in two places. After the cutover, some of the xalan-test tests might need adjustment anyway, if they make any assumptions about project structure, dependencies etc. As soon as anyone would start merging them into this project to make things easier in the future, that person would have to reconcile my tests with the migrated ones. We might even have two different test frameworks, JUnit 4/5 vs. the currently used home-brew framework of xalan-test. I guess, in a first step when migrating them into this project, we would need to run them from Surefire (unit tests) and Failsafe + Invoker (integration tests) as-is, i.e. make sure to run them without first migrating them to JUnit, if the latter would be a huge effort. My handful of new tests might even be an impediment during that phase. I can easily add those test plugins and write a handful of tests, but it is not my decision to start adding big new features to the Maven build, further stalling the cutover.

Besides, I did not add the version modules. They already existed without any test coverage. I merely ported them to the Maven build and according to the boyscout rule changed the way they work while retaining the API, because I wanted to avoid code generation magic in favour of simple resource filtering.

So, thanks for your opinion. I know you want the right thing. I am not arguing about that, we are on the same page there. I simply do not want to mix this with the Maven cutover, so our disagreement is not about "new tests yes/no" but about "new tests now/later".

@jkesselm, I know we are talking post factum, because the PR was already merged. Would you have me add test frameworks and tests rather now or later? If now, then which testing tool and version are you envisioning? My suggestion is Spock. A more conservative person would probably suggest JUnit 5 Jupiter, an even more conservative one JUnit 4.

vlsi: > You are not even working actively on the Maven branch

It is not important if I dislike Maven or not.
Please check https://www.apache.org/foundation/how-it-works/#meritocracy. I have suggested PRs for the other topics, I have answered users' questions on the mailing list.

> Earlier we talked about test automation several times, and nowhere did anyone tell me to add test plugins or dependencies before the Maven cutover

May I suggest that you should pioneer by adding the first JUnit5 test right within the PRs you suggest?
Sure the others will see how to add them and it is likely they add more.

> Are you in any position to speak authoritatively and command me to do such things?

Wow. You suggest code changes, and I suggest that you should cover your changes with tests.
Sure you can ignore my suggestions based on whatever reasons. Just be fair and don't say I am asking too much.
Previously you just said ["that is a separate (huge) PR"](https://github.com/apache/xalan-java/pull/129#discussion_r1407046638) which is either a misunderstanding (so I listed a step-by-step for adding unit test) or a deliberate lie.
Sure I never asked you to add integration tests to xalan-test-whatever, so do not put it as if "you already covered everything while I am requiring huge additional efforts".

> My suggestion is Spock

Spock is tied to Groovy, and Groovy has issues running with modern Java versions. For instance, testing with Java 22/23/.. would be problematic as in https://github.com/apache/jmeter/issues/6114#issuecomment-1825534617
I suggest avoiding Spock and Groovy unless they provide significant advantages over the regular Java/Kotlin code.

> Theoretically assuming for a minute, I would add tests separately from xalan-test, what would be the consequence?

I will send you an email in 3 years. Hopefully "maven cutover" will be merged by then, and I bet the tests will use JUnit Platform + JUnit Jupiter.

> They already existed without any test coverage. I merely ported them to the Maven build

The absence of tests should not be an excuse for not adding tests for the code you change.
It might be that innocently-looking changes introduce subtle bugs.

>I wanted to avoid code generation magic in favour of simple resource filtering

"code generation" is the same as resource filtering. I do not see why you call "replacing a variable in a file" with "magic". Sure replacing a variable is not magic.

jkesselm:
Haven't looked at Spock, or the differences between junit 4 and 5, so I don't have a valid opinion.  What's current preference/recommendation in the rest of Apache, or is there no consensus?

Apologies if I gave the impression that I was against starting to add tests. I just didn't want to deal with the complication of restructuring out test suite until we had a build which passed the existing tests and/or tests which approved the new build. If you feel like dropping in a proposed framework, to be filled in more completely later, go for it.


@vlsi, if you want to start looking at porting the test suite to a modern framework, that's a big task (in many annoying bite-sized units) that would be a significant contribution. We could probably migrate from our old processor-plugin layer to TrAX, now that it's an accepted standard, though we'd still want to have buckets which also regression test Xalan's own APIs.

Note: our test suite includes tests of W3C defined behaviors that we know we didn't implement exactly as written. Currently our drivers run these but know failures on them are not regressions when testing Xalan. We'd want to have the same capability in the new version. Addressing that cleanly might be the major design effort in this port.



--

serializer/src/main/java/org/apache/xml/serializer/Version.java
jkesselm: @vlsi: Xalan's tests currently live in the xalan-test project. I've done a quick grep there and indeed I don't see a regression test for Version. 

If you think we need one Right Now, may I suggest you write and submit one rather than asking someone else to do so? That would require either working with the existing API test driver (to be ported later), or with Maven's approach to testing (running ahead of test being moved into the main project).

Note that since this value will change over time, a naive test probably can't do much more than check for self-consistency. Checking against a reference value raises the same questions of how that value is obtained, until/unless that can be driven from the same Source Of Truth as the build. 

Though. I suppose one could check that the semantic version is equal to or more recent than the version active when this test was last modified. Personally I'd consider that, or even the bare consistency check, adequate for first draft.

kriegaex: > > You are not even working actively on the Maven branch
> 
> It is not important if I dislike Maven or not.

Read again what you just quoted. I said you are not actively working on it, but commanding around others who do.

> Please check https://www.apache.org/foundation/how-it-works/#meritocracy.

Stop lecturing me.

> > test automation (...) before the Maven cutover
> 
> May I suggest that you should pioneer by adding the first JUnit5 test right within the PRs you suggest? Sure the others will see how to add them and it is likely they add more.

Fine, done: #133. Not just unit tests but also integration tests, because if I start doing that, I want to do it right. 

> I listed a step-by-step for adding unit test

You do not need to teach me how to addor write either unit or integration tests. Like I said, I know how to do it. The debate was about whether it makes sense to add them before or after the Maven cutover. Why do I have to repeat myself so many times? Sometimes I feel like I am talking to someone suffering from amnesia.

> > My suggestion is Spock
> 
> Spock is tied to Groovy, and Groovy has issues running with modern Java versions.

That is ridiculous! I have heavily used Spock in many large-scal commercial projects, and that was never an issue

> For instance, testing with Java 22/23/..

Hilarious! The current Java release is 21, and both Groovy and Spock are under active development and regularly updated to work with current JDK versions. It will be the same here. Save your pseudo arguments for someone else, please. I want to remind you that until I personally recently added the capability to compile on JDK 9+ (including 21), this project only compiled on JDK 8. And now you start complaining about 22 or 23. If this is your biggest problem, I am happy for you, because then you do not have any real ones.

> I suggest avoiding Spock and Groovy unless they provide significant advantages over the regular Java/Kotlin code.

There are significant advantages. The BDD test DSL makes tests more readable, so they may serve as a living specification. I often discussed and designed Spock tests in the past with non-technical stakeholders, because they could understand the basic structure and I was able to code specifications (and tests) directly from their requirements. BTW: Groovy, in contrast to Kotlin, is a 99% super set of Java, i.e. every Java developer can simply start writing "Java-ish" Spock tests and learn Groovy syntax step by step. In 90% of the commercial projects I used Spock in, the code under test was written in Java, not in Groovy - only the tests were. Spock comes with a build-in mock framework, usually there is no need to use Mockito or anything else on top of it. I can express the same things with way less code, which not only saves time while writing tests but also is more readable and more easily maintainable. I know both JUnit and Spock and can compare them to each other very well from experience. Can you?

> I will send you an email in 3 years. Hopefully "maven cutover" will be merged by then

Your sarcasm is unhelpful. Even if I would not have volunteered to temporarily support Joe in his endeavour and he would be all on his own - obviously you have no intention to support him, only to crisicise everyone involved - he would be finished way sooner than that.

> I bet the tests will use JUnit Platform + JUnit Jupiter.

Probably you are right in this respect, because that is what I used in #133, despite knowing that it is a way worse choice that Spock would have been. But I do not feel so inclined to waste more of my time with fruitless debates with you, @vlsi.

> > They already existed without any test coverage. I merely ported them to the Maven build
> 
> The absence of tests should not be an excuse for not adding tests for the code you change.

I was not _excusing_ anything, I was _explaining_ something. Like I said, I would have preferred not to open another front in this war (Maven cutover PR) and rather win it battle by battle. Chances are that reviewers will complain that too many new topics are contained in one big PR in the end. But alas, now you got what you want. The whole Xalan team will have to deal with it.

> It might be that innocently-looking changes introduce subtle bugs.

I know that and agree. Again, it was not about _if_ but about _when_ to add tests.

PR #130, opened 2023-11-25
Create and populate lib directory

kriegaex:
Copy dependencies to _[root]/lib_ directory during package phase. Also Build Helper Maven Plugin to generate a property for the project root folder, using it from Maven Dependency Plugin. This can come in
handy in many situations.

@jkesselm, this should help you run the tests from _xalan-test_ against a clean Maven build, too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ target/
 
 # Xalan
 build/
+lib/
 **/xsltc/compiler/XPathLexer.java
 **/xsltc/compiler/XPathParser.java
 **/xsltc/compiler/sym.java
--- a/mvnbuild.bat
+++ b/mvnbuild.bat
@@ -1,1 +1,1 @@
-mvn clean compile site package
+mvn clean package site
--- a/mvnbuild.sh
+++ b/mvnbuild.sh
@@ -1,1 +1,1 @@
-mvn clean compile site package
+mvn clean package site
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-dependency-plugin</artifactId>
+          <version>3.6.1</version>
           <executions>
             <execution>
               <id>copy-artifact</id>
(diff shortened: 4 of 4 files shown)


jkesselm:
You removed "site" from the build script.  Remember that our "binary distributions" zipfiles need to include the documentation. AFAICT, that means site must occur before package produces the latter, which is why that order is specified in the script.

We *could* move the operations now performed in site into a phase run before package. (prepackage?). Or move generating the distribution files from package into a later one ("install" doesn't seem entirely right, but seems to come closest). Neither seems to be wholly appropriate in Maven idiom, which seems to assume site is orthogonal to other builds. 

Can we fix this? If not, the mvnbuild scripts need to be left unchanged for now.

jkesselm:
(In other words, unless I've missed something else you've done, that doesn't seem to be  just a cosmetic change.)

jkesselm:
(Note that CI on the maven build will fail until the lib/ directory *is* created.)

I'm going to merge this but put "site" back in for now -- it's a separate issue.

kriegaex:
> You removed "site" from the build script.

No, I did not. I reckon it is late at your location, maybe your eyes are just a bit tired. Please check again. I removed `compile`, because executing `package`, as explained in the commit comment, implies `compile` already. Then I simply changed the order of `package` and `site`, which should not be a problem, because they are in different build lifecycles, as I also explained. Building a Maven site is a separate lyifecycle from building and packaging the product itself. I.e., if you just run `mvn site`, you should get a complete Maven site. If you just run `mvn package`, you should get a correctly packaged product.

If a goal or phase in lifecycle A needs another one in lifecycle B, this is a smell. In our case here, if you can only successfully package after building a site, this is an error in the build configuration, because building a site is optional and therefore not part of the default lifecycle. Like I mentioned in another discussion before, each module should create its own javadocs, because the latter are needed per module to pass Maven Central upload checks anyway. If, like in this case, you want aggregated javadocs across all modules on top per-module ones and include them in the binary distro, this should be done in the build lifecycle, not in the site lifecycle (or in both, if you need it in both!). And you are right, creating javadocs needs to happen before creating the binary distro assembly, of course.

Funny detail: No matter if I run `mvn clean package site` or `mvn clean site package`, I noticed that the site (and with it the javadocs) is always generated before the assembly plugin runs, i.e. in both cases the binary distro contains the site, just as you envision. Actually, that surprises me. I would have thought that the site is created later. But because I would not rely on such behaviour, for now I understand that you reversed the order in your add-on commit to `mvn clean site package`. The real fix for that build config should get its own PR.

Another question I have is, whether you really want the full Maven site packaged into the binary distroor really just aggregated javadocs. Presently, the former happens.

jkesselm:
> If a goal or phase in lifecycle A needs another one in lifecycle B, this is a smell.

You're right, I misread. Apologies. It was tired and I was late. 

The choice to put Javadoc in site was based on examples I found online. It certainly may be absolutely the Wrong Thing for our needs. 

If I move that and the other document processing into the main flow, would compile be the recommended phase, or something else? It is, arguably, a kind of compilation... I'd have to make those depend upon the taglet module, but that's certainly doable. 

Site being a separate flow makes some sense. The problem is the established pattern that the binary distribution has included both a copy of the code and a local copy of the website, to make it a single-download "product" image. We don't _have_ to continue that pattern, if course; site can do its own "packaging"  into a separate docs zip file.

I was trying to produce "the same" zipfile content, but may have to give up - that goal and settle for equivalent. 

(As agreed, I'm staying out of your way as much as possible. I'm currently focusing on making xalan-test play nice with the Maven build. The generated lib/ directory isn't quite making it happy yet, but that's probably over-specific jarfile references that should be fixed from test's end of things. Once that's resolved I'm planning to remove the bootclasspath kluge.)

PR #132, opened 2023-11-28
Move Stylebook docs and Assembly distros generation into own profiles

kriegaex:
@jkesselm The new profiles
  - `create-docs` for Stylebook docs and
  - `create-distros` for source and binary assemblies

are both active by default, but can be deactivated on the Maven command line or from an IDE to achieve faster development builds, e.g. if in the future a developer just wants to change some source code, compile and test it. He does not need any documentation or distros for that, so he can just build with something like:

```text
mvn [goal|phase] -P !create-docs,!create-distros
```

Relates to https://issues.apache.org/jira/browse/XALANJ-2708.

P.S.: I think, the fact that the Xalan JAR (not the distro, just the single module) contains dependency classes is not nice. When using Xalan as a library, it would be better to let Maven or Gradle pull the transitive dependencies. To replicate the old behaviour, we have to use Maven Shade, which is not just slow, but IMO also replicates an old Xalan problem: The dependency classes are included with their original package names, i.e. `org.apache.bcel`. If including them at all would make any sense, they should be relocatedin order to avoid clashes for users who happen to use e.g. BCEL in their own applications, possibly in another version. Classpath conflics are pre-programmed like this. But I guess, that issue deserves its own issue and its own discussion.


jkesselm:
> P.S.: I think, the fact that the Xalan JAR (not the distro, just the single module) contains dependency classes is not nice. 

I had done some work to eliminate that. If they have crept back in, I agree we should pull them out again. 

This is one of the things I was (ab?)using "provided" for; it was the simplest way I had found to say "this component will be provided separately, so don't copy it into the jarfiles". 

If there's a better way to express that, _great_. What incantation have I not learned yet?

kriegaex:
It is not so much about anything new having crept in but about the ones which have been part of the Xalan library in previous (Ant) releases already. I am assuming that was done intentionally.

Let us continue the discussion in https://issues.apache.org/jira/browse/XALANJ-2713 and try to stay focused on the PR here.

PR #133, opened 2023-11-29
Add unit and integration testing infrastructure and initial tests for version classes

kriegaex:
We have a full set of parametrised unit tests as well as 3 integration tests here, all based on JUnit 5 Jupiter (and Mockito for the unit tests), despite me having a JUnit allergy. The tests are so much more tedious to write than Spock tests and still look so much uglier. I have introduced Spock to dozens of teams as an agile coach in the last 10+ years, and all of them never looked back to JUnit once they got infected and had worked with it for 2 weeks. It causes me almost physical pain to use JUnit, especially the nightmarish syntax for parametrisation. But in contrast to some other people who think it is a "waste of time" to work with a tool they dislike, I am trying to contribute something anyway.

Please read the commit comments for more info or just review the changes.

You can run the unit tests by `mvn clean test`, the integration tests + unit tests by `mvn clean verify`, only the ITs by `mvn -DskipTests clean verify`. If you want to `verify`, `install` or `deploy` and skip the ITs, simply deactivate their profile, using e.g. `mvn -P !integration-tests clean install`.

@jkesselm, please note that this PR was committed on top of the commit for #132, i.e. it is a super set of it. I wanted to avoid possible merge conflicts later, so my suggestion is to review #132 and possibly merge first, so I can rebase this one, or to just merge this compound PR to cover both PRs at once. If you would fast-forward instead of merge, it would be the same anyway, it only makes a difference if you create actual merge commits.


kriegaex:
@jkesselm already knows, but for anyone else reading this, such as @vlsi, because he urged me to add test infrastructure to the project: I downgraded from Mockito 5.x to 4.x to cater to the requirement to be able to build on JDK 8 (Mockito 5 requires JDK 11+). I also managed to weed out a platform dependency in one of the ITs. Now the CI build `clean verify` passes. The overall workflow only fails, because a later build step with running [_xalan-test_](https://github.com/apache/xalan-test) tests is known to fail and needs fixing.

vlsi:
This looks great to me, however, I am not sure it is good to use Mockito here. Sure it would be a great source for copy-pasting the next steps, however, mocking here might be a little bit of overkill.

I would rather extract the "version parsing method" to a separate class (e.g. with `String` constructor). Then the parsing itself could be tested without mocks, and the end-to-end could be tested with, well, integration.

In other words, I would trade "slightly less test coverage" for "less maintenance overhead of dealing with Mockito tests/upgrades vs Java versions".

---

>I downgraded from Mockito 5.x to 4.x to cater to the requirement to be able to build on JDK 8

An alternative solution could be skipping some of the tests when running with JDK 8.
For instance, you could exclude the test file that uses Mockito with `<testExcludes>` and be done with it: https://github.com/pgjdbc/pgjdbc/blob/6f11a2854ef67aec16dd413e2fa88bd8512eaf3d/pgjdbc/reduced-pom.xml#L138-L145

---

> Now, the CI build clean verify passes

It would be great to actually see the CI results, and it would be great to see CI on both Windows and Linux machines as I expect there might be newline/pathseparator-like issues

jkesselm:
> It would be great to actually see the CI results

Should now be visible. For security reasons, CI doesn't get run on PRs until a commiter allows them to run.

> and it would be great to see CI on both Windows and Linux machines

You drafted the YAML for CI. If you know how to request both runs, you could submit a change for it. I believe CI is now present on the main branch, so your PR should probably be against that and theoretically we should be able to pull into the maven branch via rebase.

vlsi:
Here you go: https://github.com/apache/xalan-java/pull/136

jkesselm:
kriegaex: #132 has been merged. Let me know when this one has been rebased.

kriegaex:
@jkesselm, rebased on latest base branch.

kriegaex:
> I would trade "slightly less test coverage" for "less maintenance overhead of dealing with Mockito tests/upgrades vs Java versions".

I would not. The right thing to do would be to
  * introduce a `xalan-commons` module (there is a lot more duplicate code in Xalan vs. Serializer modules),
  * put a singleton-style versions base class with non-static methods (easily testable) there, the static ones only being facades for calling the singleton instance's ones,
  * making sure that tests can inject a testable instance.

But that is for another day. Static methods or not, a proper unit test should cover all code and stub out file I/O, i.e. one way or another we end up with a partial mock, if we want clean in-memory tests. Especially with parametrised tests, I do not wish to store a whole bunch of test resource files for different test cases. Trading coverage for slightly simpler tests is an option, but not the one I chose for good reason. 

> > I downgraded from Mockito 5.x to 4.x to cater to the requirement to be able to build on JDK 8
> 
> An alternative solution could be skipping some of the tests when running with JDK 8.

I prefer my alternative, not skipping tests. We would have to skip **all** tests using mockito on JDK 8 then, but good unit tests ought to test the unit under test in isolation, i.e. I want to use mocks in many situations. This class is not complex, but others are and have dependencies that I want to mock when injecting them.

@jkesselm, mocking might be over-used, but it does not mean that I do. Mocks or test doubles in general are important tools for good unit tests.

> > Now, the CI build clean verify passes
> 
> It would be great to actually see the CI results

I checked them, so they were visible already. Because of the strange policy that each single CI run has to be manually approved by a committer, I simply checked the results on my fork, from where I started the PR. There, the tests run automatically. You can look there, too. In case of your own PIs, that is where you can also verify the build results, if you are not a committer (I have no idea, actually).

> and it would be great to see CI on both Windows and Linux machines as I expect there might be newline/pathseparator-like issues

I agree, but it was way beyond the scope of this PR to add a build matrix here. Actually, OS matrix is not the only thing. We also want a JDK version matrix. I suggest 8, 11, 17, 21.

jkesselm:
> Where was I testing code against a copy of itself?

_You weren't_. Vlad's suggestion of avoiding mockito by having the test do its own read and parse of version appeared to be essentially that, unless I misunderstood his suggestion.

kriegaex:
> > Where was I testing code against a copy of itself?
> 
> _You weren't_. Vlsi's suggestion of avoiding mockito by having the test access the version parsing code directly ([#133 (comment)](https://github.com/apache/xalan-java/pull/133#issuecomment-1833170148)) appeared to be essentially that, unless I misunderstood his intent. Which I think I did. Withdrawing comment.

Oh, OK. Thanks for the clarification. In Vladimir's defence, I do not think he suggested that either. 🙂

kriegaex:
FYI, I added another small commit aff65011. At first, I wanted to squash it into the commit adding the ITs, because it does the same thing for Maven Shade which there I also did for other plugins. But being unsure if anyone is in the middle of reviewing something, I wanted to avoid squash & force push for now. I can still squash before the final merge.

xalan/src/main/java/org/apache/xalan/Version.java
```diff
@@ -51,12 +49,24 @@ public class Version
...
+    try (InputStream fromResource = getPropertiesStream()) {
+      Properties properties = new Properties();
+      if (fromResource != null) {
+        properties.load(fromResource);
+        return properties.getProperty("version", NO_VERSION);
+      }
+    } catch (Exception e) {
+      new RuntimeException("Cannot read properties file to extract Xalan version number information: ", e)
+        .printStackTrace();
+    }
```
vlsi, line 65: It might be a slightly better idea to use `catch(Throwable)` since `Exception` does not cover all the exception types.

xalan/src/test/java/org/apache/xalan/VersionTest.java
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+
+  @Test
+  public void testCannotReadProperties() {
+    try (MockedStatic<Version> versionMock = Mockito.mockStatic(Version.class, Mockito.CALLS_REAL_METHODS)) {
+      versionMock
+        .when(Version::getPropertiesStream)
+        .thenThrow(NullPointerException.class);
+      assertEquals("0.0.0", Version.readVersionNumber());
```
vlsi, line 61: Do you expect NPE from `getPropertiesStream`? The method is a single line only (~`.getResourceAsStream`), and Javadoc says the method returns `null` in case the value is not found. Is there a real value in testing "what happens if getPropertiesStream throws NPE"?

vlsi:
> We would have to skip all tests using mockito on JDK 8 then

Right you are. Of course, it is too early to predict how many Mockito-using tests will be there in Xalan. At the same time, I estimate there will be a sheer amount of tests that do not use mocks, so they would be pretty much runnable with JDK8.

In the current case, the same test coverage can be achieved without Mockito (see below), and it would not require much effort.

Yet another possibility might be having different test dependency versions for different JDK versions. That is, of course, a weak idea for API-rich libraries, however, it might still allow running some of the tests with multiple JDKs.

> The version parsing method already is separate, and there is a separate test for it, which does not need mocks. I.e., I already did what you suggested before you even asked for it. Please check again

You are right. Let me clarify what I meant.

Currently, you have `static void parseVersionNumber(String version) {`, and you have `VersionAccessor` to access that method.
Sure it works, and I have no issues with that.

I meant a slightly different design (sure it would be generated boilerplate with Java 8, but it does not change the overall picture):

```java
/* non-public or semi-public */ record XalanVersionRecord(
  int major, int release, int maintenance,
  int development,
  boolean snapshot
) {
    XalanVersionRecord(String versionString) {
        this(versionString.parseIntoVersionFields());
    }
}
```

`Xalan.java` would be like `public static int getMajorVersionNum() { return INSTANCE.major(); }`
In other words, you keep the old class like it is with its static methods, however, you could move all the parsing logic to a regular non-static class, so you can have nice final fields there, and you can have nice constructors.

I think having a struct with all final fields is a little bit easier to comprehend than having multiple mutable static fields + `static void resetVersionNumber` and things like that.

If you add `XalanVersionRecord(InputStream ..)` constructor, then you no longer need mocks in `testReadProperties`.
I see you have a test that verifies "NPE from getPropertiesStream", however, frankly speaking, I think the test does not add much to the test coverage.

Sure the current code works well.

Of course, you can add `InputStream` parameter to your current `static readVersionNumber` method, and it would eliminate the need for the mock.

Of course, there are cases when mocks are helpful. I just mean in the current case, Mockito does add much, and, well, adding more complicated mocks would be a definite overkill for testing versions.

----

@kriegaex , the Maven part of the build prints 5900 messages which makes it hard to scroll and understand. 
What is the use of messages like "Generating /home/runner/work/xalan-java/xalan-java/target/site/apidocs"?
What is the purpose of messages like "Loading source files for package org.apache.xalan.res"?
What is the purpose of "Warning:  Entry: xalan-project_2.7.3/docs/apidocs/org/apache/xml/serializer/utils/class-use/SerializerMessages_zh_TW.html longer than 100 characters."? Sure file paths can exceed 100 characters, and I do not see why print such a message.

CI prints a stack trace even though the tests pass: https://github.com/apache/xalan-java/actions/runs/7055388836/job/19206140606?pr=133#step:4:2885
Can you somehow make it avoid printing the error in case the exception was intentional?
Of course, we want negative tests, however, I doubt that we want to print stack traces in case everything works smoothly.

jkesselm:
> 5900 messages

I'm away from keyboard, but according to the docs, simply adding the -q (quiet) option to the mvn command line should silence all the info stuff and maybe the warnings, outputting only error messages.

I've sometimes found the usual stream of progress stuff and "this might be cleaner"  useful during development, but I agree that CI can be terser. Of course humans can specify -q or not as they prefer.

vlsi:
> -q (quiet) option to the mvn command line should silence all the info stuff and maybe the warnings, outputting only error messages

It might be silencing too much though (it cuts every Maven output, so even "build succeeded" is not there which is strange for a day-to-day use). `-q` does not silence stack traces from the test. I think the tests should either redirect `System.err` to a different stream or use a logger (e.g. JUL) to allow users to control the verbosity.

---

Running the tests with Java 22 fails due to Groovy not supporting Java 22.

```
% java -version
openjdk version "22-ea" 2024-03-19
OpenJDK Runtime Environment (build 22-ea+25-1998)
OpenJDK 64-Bit Server VM (build 22-ea+25-1998, mixed mode, sharing)
```

```
Running post-build script: /.../xalan-java/integration-tests/target/it/VersionCheck_JarsOnBootClasspath/verify.groovy
BUG! exception in phase 'semantic analysis' in source unit 'Script1.groovy' Unsupported class file major version 66
	at org.codehaus.groovy.control.CompilationUnit$ISourceUnitOperation.doPhaseOperation(CompilationUnit.java:901)
	at org.codehaus.groovy.control.CompilationUnit.processPhaseOperations(CompilationUnit.java:692)
	at org.codehaus.groovy.control.CompilationUnit.compile(CompilationUnit.java:666)
	at groovy.lang.GroovyClassLoader.doParseClass(GroovyClassLoader.java:373)
	at groovy.lang.GroovyClassLoader.lambda$parseClass$2(GroovyClassLoader.java:316)
	at org.codehaus.groovy.runtime.memoize.StampedCommonCache.compute(StampedCommonCache.java:163)
	at org.codehaus.groovy.runtime.memoize.StampedCommonCache.getAndPut(StampedCommonCache.java:154)
	at groovy.lang.GroovyClassLoader.parseClass(GroovyClassLoader.java:314)
	at groovy.lang.GroovyShell.parseClass(GroovyShell.java:572)
	at groovy.lang.GroovyShell.parse(GroovyShell.java:585)
	at groovy.lang.GroovyShell.parse(GroovyShell.java:597)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:459)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:495)
	at groovy.lang.GroovyShell.evaluate(GroovyShell.java:469)
	at org.apache.maven.shared.scriptinterpreter.GroovyScriptInterpreter.evaluateScript(GroovyScriptInterpreter.java:69)
	at org.apache.maven.shared.scriptinterpreter.ScriptRunner.executeRun(ScriptRunner.java:223)
	at org.apache.maven.shared.scriptinterpreter.ScriptRunner.run(ScriptRunner.java:160)
	at org.apache.maven.plugins.invoker.AbstractInvokerMojo.runPostBuildHook(AbstractInvokerMojo.java:1902)
	at org.apache.maven.plugins.invoker.AbstractInvokerMojo.runBuild(AbstractInvokerMojo.java:1883)
	at org.apache.maven.plugins.invoker.AbstractInvokerMojo.runBuild(AbstractInvokerMojo.java:1542)
	at org.apache.maven.plugins.invoker.AbstractInvokerMojo.lambda$runBuilds$4(AbstractInvokerMojo.java:1304)
	at org.apache.maven.plugins.invoker.JobExecutor.lambda$null$1(JobExecutor.java:62)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1570)
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 66
	at groovyjarjarasm.asm.ClassReader.<init>(ClassReader.java:199)
	at groovyjarjarasm.asm.ClassReader.<init>(ClassReader.java:180)
	at groovyjarjarasm.asm.ClassReader.<init>(ClassReader.java:166)
	at groovyjarjarasm.asm.ClassReader.<init>(ClassReader.java:287)
	at org.codehaus.groovy.ast.decompiled.AsmDecompiler.parseClass(AsmDecompiler.java:83)
	at org.codehaus.groovy.control.ClassNodeResolver.findDecompiled(ClassNodeResolver.java:255)
	at org.codehaus.groovy.control.ClassNodeResolver.tryAsLoaderClassOrScript(ClassNodeResolver.java:193)
	at org.codehaus.groovy.control.ClassNodeResolver.findClassNode(ClassNodeResolver.java:175)
	at org.codehaus.groovy.control.ClassNodeResolver.resolveName(ClassNodeResolver.java:129)
	at org.codehaus.groovy.control.ResolveVisitor.resolveToOuter(ResolveVisitor.java:772)
	at org.codehaus.groovy.control.ResolveVisitor.resolve(ResolveVisitor.java:413)
	at org.codehaus.groovy.control.ResolveVisitor.resolveFromDefaultImports(ResolveVisitor.java:567)
	at org.codehaus.groovy.control.ResolveVisitor.resolveFromDefaultImports(ResolveVisitor.java:531)
	at org.codehaus.groovy.control.ResolveVisitor.resolve(ResolveVisitor.java:412)
	at org.codehaus.groovy.control.ResolveVisitor.resolve(ResolveVisitor.java:374)
	at org.codehaus.groovy.control.ResolveVisitor.resolveOrFail(ResolveVisitor.java:342)
	at org.codehaus.groovy.control.ResolveVisitor.resolveOrFail(ResolveVisitor.java:333)
	at org.codehaus.groovy.control.ResolveVisitor.resolveOrFail(ResolveVisitor.java:329)
	at org.codehaus.groovy.control.ResolveVisitor.transformConstructorCallExpression(ResolveVisitor.java:1125)
	at org.codehaus.groovy.control.ResolveVisitor.transform(ResolveVisitor.java:803)
	at org.codehaus.groovy.control.ResolveVisitor.transformMethodCallExpression(ResolveVisitor.java:1141)
	at org.codehaus.groovy.control.ResolveVisitor.transform(ResolveVisitor.java:799)
	at org.codehaus.groovy.control.ResolveVisitor.transformDeclarationExpression(ResolveVisitor.java:1165)
	at org.codehaus.groovy.control.ResolveVisitor.transform(ResolveVisitor.java:795)
	at org.codehaus.groovy.ast.ClassCodeExpressionTransformer.visitExpressionStatement(ClassCodeExpressionTransformer.java:129)
	at org.codehaus.groovy.ast.stmt.ExpressionStatement.visit(ExpressionStatement.java:41)
	at org.codehaus.groovy.ast.CodeVisitorSupport.visitBlockStatement(CodeVisitorSupport.java:86)
	at org.codehaus.groovy.ast.ClassCodeVisitorSupport.visitBlockStatement(ClassCodeVisitorSupport.java:168)
	at org.codehaus.groovy.control.ResolveVisitor.visitBlockStatement(ResolveVisitor.java:1388)
	at org.codehaus.groovy.ast.stmt.BlockStatement.visit(BlockStatement.java:72)
	at org.codehaus.groovy.ast.ClassCodeVisitorSupport.visitClassCodeContainer(ClassCodeVisitorSupport.java:142)
	at org.codehaus.groovy.ast.ClassCodeVisitorSupport.visitConstructorOrMethod(ClassCodeVisitorSupport.java:115)
	at org.codehaus.groovy.ast.ClassCodeExpressionTransformer.visitConstructorOrMethod(ClassCodeExpressionTransformer.java:87)
	at org.codehaus.groovy.control.ResolveVisitor.visitConstructorOrMethod(ResolveVisitor.java:297)
	at org.codehaus.groovy.ast.ClassCodeVisitorSupport.visitMethod(ClassCodeVisitorSupport.java:110)
	at org.codehaus.groovy.control.ResolveVisitor.visitMethod(ResolveVisitor.java:268)
	at org.codehaus.groovy.ast.ClassNode.visitMethods(ClassNode.java:1129)
	at org.codehaus.groovy.ast.ClassNode.visitContents(ClassNode.java:1122)
	at org.codehaus.groovy.ast.ClassCodeVisitorSupport.visitClass(ClassCodeVisitorSupport.java:52)
	at org.codehaus.groovy.control.ResolveVisitor.visitClass(ResolveVisitor.java:1334)
	at org.codehaus.groovy.control.ResolveVisitor.startResolving(ResolveVisitor.java:258)
	at org.codehaus.groovy.control.CompilationUnit.lambda$addPhaseOperations$4(CompilationUnit.java:218)
	at org.codehaus.groovy.control.CompilationUnit$ISourceUnitOperation.doPhaseOperation(CompilationUnit.java:897)
	... 25 more
*** end build.log for: VersionCheck_JarsOnBootClasspath/pom.xml ***

[ERROR] -------------------------------------------------
[ERROR]
[ERROR] The following builds failed:
[ERROR] *  VersionCheck_JarsOnClasspath/pom.xml
[ERROR] *  VersionCheck_ShadedJarOnBootClasspath/pom.xml
[ERROR] *  VersionCheck_JarsOnBootClasspath/pom.xml
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-invoker-plugin:3.6.0:run (integration-test) on project integration-tests: 3 builds failed. See console output above for details. -> [Help 1]
[ERROR]
```

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: `Exception` does cover all exception types, just not all throwable types. For instance, `Error`s ought not be caught. They are supposed to be escalated and ought to exit the application, which is why I am only catching exceptions.

xalan/src/test/java/org/apache/xalan/VersionTest.java
vlsi: > I do not expect a NPE, but the javadoc says that NPE will be thrown if the input parameter is null

The input parameter is `"version.properties"` string literal, so NPE must not happen.
Frankly, testing NPE there is overkill.

At the same time, "IO" exception occurs **while accessing the stream** rather than at `getResourceAsStream`, so if you want to test how it behaves in case the stream misbehaves, then you'd better have a stream that throws IO exception from its `read` call.

xalan/src/main/java/org/apache/xalan/Version.java
vlsi: Technically, Java calls them "exceptions", not "throwables": https://docs.oracle.com/javase/specs/jls/se21/html/jls-11.html

> **Every exception** is represented by an instance of the class **`Throwable`** or one of its subclasses ([§11.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-11.html#jls-11.1))

kriegaex:
> In the current case, the same test coverage can be achieved without Mockito (see below)

Not true. Did you even read myprevious reply? The parsing logic is not using any mocks already. The only mock used is to stub out the I/O operation, and for that the isolated staic method mock is necessary. I do not want any I/O in a unit test. Your example is irrelevant, because it targets the wrong method. Of course I could factor out that class, I even was the first one to suggest it. But I am not doing it, because that code is nicely covered without any mocks already. Over-engineering. I had absolutely no problem to get 100% test coverage for this class without any complicated test code. The test is straightforward.

> Of course, there are cases when mocks are helpful. I just mean in the current case, Mockito does add much

It adds I/O isolation. Valuable enough for me.

> adding more complicated mocks would be a definite overkill for testing versions.

Nothing is complicated here. If any developer thinks that this mock usage is complicated, he has not seen any complicated mocks yet.

> the Maven part of the build prints 5900 messages which makes it hard to scroll and understand. Can you somehow adjust the configuration so it does not print that many noisy messages?

If anything goes wrong in the build, I want to see the details. Won't do.

> CI prints a stack trace even though the tests pass: https://github.com/apache/xalan-java/actions/runs/7055388836/job/19206140606?pr=133#step:4:2885 Can you somehow make it avoid printing the error in case the exception was intentional?

It is meant to print that stack trace,because that is the defined behaviour. I want to see it, because the test is verifying that error condition.

kriegaex:
> Running the tests with Java 22 fails due to Groovy not supporting Java 22.

Nitpicking. The build is not meant to run on JDK 22. I already explained.

xalan/src/test/java/org/apache/xalan/VersionTest.java
vlsi: Just in case, here's an `InputStream` that throws:

```java
new InputStream() {
  @Override
  public int read() throws IOException {
    throw new IOException("not implemented");
  }
};
```

In other words, testing IO failure here with regular Java is not even harder than testing with Mockito.

kriegaex:
> I think the tests should either redirect `System.err` to a different stream

No, not thread-safe. More important sysErr output might get suppressed, too.

> or use a logger (e.g. JUL) to allow users to control the verbosity

You keep repeating the same suggestions, even though I already explained elsewherewhy I am not going to do this for that little version class: Xalan-J currently does not use any logging frameworks, and I am not going to introduce one for a single class. That is way out of scope and a big decision to be made by the maintainers, not by a me.

I see that you have not identified any real problems with this PR, because you focus on miniscule details. That makes me happy, because it indicated that my code is quite OK. 🙂 Thanks for all your feedback. I know you mean well. I just disagree.

kriegaex:
> I agree that CI can be terser

The opposite is true IMO. As we have no way to remote-debug CI,the only chance we have to identify problems there is verbose logging. Not debug level, of course, but default Maven logging.

xalan/src/test/java/org/apache/xalan/VersionTest.java
kriegaex: > The input parameter is `"version.properties"` string literal, so NPE must not happen. Frankly, testing NPE there is overkill.

Please read more carefully. I said the NPE in the test is just a stand-in for **any** runtime exception which could happen there. It was just an arbitrary choice, because it does not matterwhich exception I am throwing,as all exceptions are cought. So we are covering the catch block, which is the objective here. That should be obvious enough.

kriegaex: > At the same time, "IO" exception occurs **while accessing the stream** rather than at `getResourceAsStream`, 

Again, read more carefully. I was talking about I/O exception due to using try with resources.

image

> so if you want to test how it behaves in case the stream misbehaves

I am testing anything that can go wrong in the whole try-block including the statement in parentheses right on top.

kriegaex: > Just in case, here's an `InputStream` that throws

Thanks, irrelevant again.

> In other words, testing IO failure here with regular Java is not even harder than testing with Mockito.

Wrong. I/O happens in method Version::getPropertiesStream, and there I cannot inject anything unless I factor out that one line of code again, which is a recursive problem. At the end, I need to stub that method if I want to avoid real I/O to happen in the test, which is exactly my objective: I do not want the test to read any test resource file but run the test in memory. The by far simplest method is to simply stub that one-line method. Is that so hard to understand? It is almost too trivial to explain IMO.

xalan/src/main/java/org/apache/xalan/Version.java
kriegaex: Stop nitpicking obsessively, won't you? You explicitly asked me to catch `Throwable`, I explained why I won't, because that would encompass fatal `Error` types.

jkesselm:
> > I agree that CI can be terser
> 
> The opposite is true IMO. As we have no way to remote-debug CI, the only chance we have to identify problems there is verbose logging. Not debug level, of course, but default Maven logging.

I split the difference. Default logging includes a lot of stuff that we really don't need to see every time right now, such as the javadoc warnings (best I was able to do was make them not be errors). Yes, the right answer is for someone to find time to clean up the javadoc at least enough to silence these complaints, but we don't need to know that during CI. 

Might be able to use grep to filter out [warning] lines, if the CI system has a portable way if doing that. If we need that any point, we can always check in a change to remove that filter temporarily.

If I have to re-run a test locally for analysis after CI says there's an issue, I'm willing to do so. What I want from CI is usually an alert, not diagnosis. With the uncommon exception of transient problems that are hard to reproduce, or environments I can't replicate. But I may be out of date here.

kriegaex:
> I say: "This looks great to me". I was OK merging this from the beginning.

I read that. But you did not say it was OK to merge. And you wrote that sentence **before** you started flooding me with review comments indicating otherwise.

> here's a diff that replaces Mockito with a line of Java code keeping the same test coverage. Am I missing something?

Yes, you are missing that test coverage is not the whole story here. You pulled the `getPropertiesStream()` call out of the method handling possible I/O (or other runtime) exceptions. I.e., if anything goew wrong during that I/O operatione, e.g. because our build is broken and the resource file is missing in the JAR, then we have an unhandled exception directly when loading the class. My solution keeps Xalan running and returning version 0.0.0, even if something goes wrong there. It is not worth killing a whole application with an uncaught error for a minor tool class.

xalan/src/test/java/org/apache/xalan/VersionTest.java
jkesselm: We're getting into a bad pattern of trying to design in reviews. (I've been guilty of it too.)

May I suggest that we keep review focused on "is it an improvement" and "is there any reason _not_ to commit", and move extended discussion back to the dev list?

That will let us merge sooner and with less noise. One can always submit another PR later if we agree there's an improvement.

Don't let the quest for perfect drive out what's good enough for now. Incremental steps.

kriegaex:
> > The opposite is true IMO. As we have no way to remote-debug CI, the only chance we have to identify problems there is verbose logging. Not debug level, of course, but default Maven logging.
> 
> I split the difference. Default logging includes a lot of stuff that we really don't need to see every time (...)

If the build passes, nobody will read the full log all the time anyway. If it fails, the error(s) will be towards the bottom or otherwise easily identifiable in the log. Question: Do the log lines "eat any bread", as we say in German? They do not cost us either time or local paid storage. Is there even a real problem to solve here? I consider it a luxury to have morelogs than we might need. But actually, I appreciate that the javadoc logs keep nagging use, because that incentivises us to either fix the warnings or find another way to handle the situation. I love information radiators,especially the ones I get for free. Usually the test passes, and then the green dot is all I need to know. If itfails, I am happy to get my hands on all info I can possible get.

vlsi:
> You pulled the getPropertiesStream() call out of the method handling possible I/O (or other runtime) exceptions. I.e., if anything goew wrong during that I/O operation,

>e.g. because our build is broken and the resource file is missing in the JAR

If the resource is missing, then `getResourceAsStream` returns `null`, and there's a test for that.

> then we have an unhandled exception directly when loading the class

You seem to assume `Errors` (e.g. `VerifyError`) should not be caught, so they would crash the app anyway.

> My solution keeps Xalan running and returning version 0.0.0, even if something goes wrong there. It is not worth killing a whole application with an uncaught error for a minor tool class

The same happens with my solution as well.

Just in case, here's an execution of my test.
The test succeeds (it ensures the versions is `0.0.0`), and the exception is reported with `System.err` just like with your changes.

Please check again, I am sure I missed nothing.

```
java.lang.RuntimeException: Cannot read properties file to extract Xalan version number information: 
	at org.apache.xalan.Version.readVersionNumber(Version.java:63)
	at org.apache.xalan.VersionTest.testCannotReadProperties(VersionTest.java:54)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:728)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.lambda$invoke$0(InterceptingExecutableInvoker.java:93)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$InterceptedInvocation.proceed(InvocationInterceptorChain.java:106)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.proceed(InvocationInterceptorChain.java:64)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.chainAndInvoke(InvocationInterceptorChain.java:45)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain.invoke(InvocationInterceptorChain.java:37)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:92)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker.invoke(InterceptingExecutableInvoker.java:86)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$7(TestMethodTestDescriptor.java:218)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.invokeTestMethod(TestMethodTestDescriptor.java:214)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:139)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.execute(TestMethodTestDescriptor.java:69)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:151)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.invokeAll(SameThreadHierarchicalTestExecutorService.java:41)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$6(NodeTestTask.java:155)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$8(NodeTestTask.java:141)
	at org.junit.platform.engine.support.hierarchical.Node.around(Node.java:137)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$executeRecursively$9(NodeTestTask.java:139)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.executeRecursively(NodeTestTask.java:138)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.execute(NodeTestTask.java:95)
	at org.junit.platform.engine.support.hierarchical.SameThreadHierarchicalTestExecutorService.submit(SameThreadHierarchicalTestExecutorService.java:35)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutor.execute(HierarchicalTestExecutor.java:57)
	at org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine.execute(HierarchicalTestEngine.java:54)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:198)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:169)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:93)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.lambda$execute$0(EngineExecutionOrchestrator.java:58)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.withInterceptedStreams(EngineExecutionOrchestrator.java:141)
	at org.junit.platform.launcher.core.EngineExecutionOrchestrator.execute(EngineExecutionOrchestrator.java:57)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:103)
	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:85)
	at org.junit.platform.launcher.core.DelegatingLauncher.execute(DelegatingLauncher.java:47)
	at org.junit.platform.launcher.core.SessionPerRequestLauncher.execute(SessionPerRequestLauncher.java:63)
	at com.intellij.junit5.JUnit5IdeaTestRunner.startRunnerWithArgs(JUnit5IdeaTestRunner.java:57)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:232)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:55)
Caused by: java.io.IOException: IO exception to test how version parser behaves to IO Exceptions
	at org.apache.xalan.VersionTest$1.read(VersionTest.java:58)
	at java.base/java.io.InputStream.read(InputStream.java:284)
	at java.base/java.io.InputStream.read(InputStream.java:217)
	at java.base/java.util.Properties$LineReader.readLine(Properties.java:504)
	at java.base/java.util.Properties.load0(Properties.java:420)
	at java.base/java.util.Properties.load(Properties.java:409)
	at org.apache.xalan.Version.readVersionNumber(Version.java:59)
	... 69 more
```

, then we have an unhandled exception directly when loading the class

@kriegaex , Please

kriegaex:
> If the resource is missing, then `getResourceAsStream` returns `null`, and there's a test for that.

Conceded. Anyway, it is an I/O operation, and something can go wrong there, if not in the specific JDK version you are currently looking at, then in another. We know how long-lived old Xalan code was. We are talking decades here, not just years. I do not see any reasonto take any unnecessary risks here, putting an unprotected I/O operation into a static block. It is called defensive programming. Maybe you heard the term. I am not going to change the application code, hence also not the test covering same code.

xalan/src/test/java/org/apache/xalan/VersionTest.java
kriegaex: > We're getting into a bad pattern of trying to design in reviews. (I've been guilty of it too.)
> 
> May I suggest that we keep review focused on "is it an improvement" and "is there any reason _not_ to commit", and move extended discussion back to the dev list?

Yes, **pleeease!**

Joe, do you remember my post on the mailing list about dysfunctional, micro-managing code reviews in OSS projects? This is exactly what I was talking about.

vlsi:
> Question: Do the log lines "eat any bread", as we say in German?

They do. Frankly, if CI fails (e.g. due to a failed test), I typically scroll the log to check the failure. The current 5900 lines are hard to scroll even with my quite powerful notebook. I wish the lines like "Generating /home/runner/work/xalan-java/xalan-java/target/site/apidocs" were never there. They are never helpful.

---

> Anyway, it is an I/O operation, and something can go wrong there

Please, IO operations throw `IOException`.
`Class.getResourceAsStream` is not an IO operation. It is a request to fetch the resource, and it is up to classloader if the operation is IO or not.

The documentation (==specification) for `getResourceAsStream` is quite clear that is shoult not throw exceptions, so discussing "unexpected exceptions from getResourceAsStream" does not sound right to me. Remember: we are not reading the stream there. We just locate it. It should either return or return null. That is it.

kriegaex:
> > Question: Do the log lines "eat any bread", as we say in German?
> 
> They do. Frankly, if CI fails (e.g. due to a failed test), I typically scroll the log to check the failure. The current 5900 lines are hard to scroll even with my quite powerful notebook.

Navigating logs is also a developer skill. I never had any big problems doing so. Sure, 5900 lines are not little, but also not much. Maybe you want to select the option to open the raw log without fancy HTML navigation and syntax colouring instead. It scrolls much faster.

Either way, this is totally out of scaope of this PR. The extensive logs you complain about do not come frommy changes but from stuff that was put in place by someone else and exists without this PR, too. Most OSS project using Maven Invoker

1. stream the fulllogs of the embedded Maven builds into their main build logs and
2. also activate debug logging for each IT build.

I deactivated both already. So, I am considerate about this issue. The passing ITs produce next to no logs (only in case of an IT failure, the build log for that IT is being streamed), the unit tests not much, only the logged exceptions for the reproduced error cases, which are helpful context info.

kriegaex:
> > It is called defensive programming
> 
> In any case, if you want to be extremely defensive (...)

Is this nitpicking some kind of addiction for you? Can you please do something productive instead? Thanks, much obliged.

vlsi:
> Is this nitpicking some kind of addiction for you? Can you please do something productive instead? Thanks, much obliged. Just save your energy for reviews of actually bad code, maybe.

@kriegaex , first, I truly do not like when people accuse me or say in public that my words are "not true". Sure I can live with that, however, I would appreciate it if you withdraw your "not true". Sure I can make mistakes, and it would be great to learn what I miss. However, in this case, I am extremely confident that my claim was true. Thank you.

Currently, it looks like you accuse me of spreading false information and causing spam of messages while the core of the issue is that it was you who claimed "not true".

Second, I started this with "This looks great to me". If you watch for a literal pattern "this is good to merge", ok, I can add it every time, however, I was 100% sure my first comment was already like "fine, there are small comments".

---

> The extensive logs you complain about do not come frommy changes but from stuff that was put in place by someone else and exists without this PR, too

@kriegaex , please consider that stacktraces with `java.lang.RuntimeException: Cannot read properties file to extract Xalan version number information: ` were added **by the current PR**. Those stacktraces will easily be misleading as it would take time to tell if the stacktrace comes from a true failure or if the stacktrace comes from a test that just happens to mock a failure.
Do not get me wrong, but it adds ~200 lines to the log output for the stacktraces alone (one for xalan, and another for serializer module).

I did mention that stacktraces from successful test exceptions are unhelpful and distracting, and I do not see why you suggest "open the raw log without fancy HTML" instead of tuning down the unwanted messages. There's no point in printing exceptions in case the test was successful.

>only the logged exceptions for the reproduced error cases, which are helpful context info.

Please, they are not helpful. They are distracting.
Imagine we will have more tests in the future. Imagine we have 50 exceptions printed (e.g. because we test positive and negative cases). Imagine one of those 50 exceptions is a new unexpected failure. How would you identify the true failure from the ones that come from the successful executions?

Please, can we avoid polluting the log?

kriegaex:
> If this is not an expected exception, I would have expected the test to report a failure.

But it **is** an expected exception. Covering the code handling this exception is the whole point of the negative test. Hence, the exception to be logged is exactly what is supposed to happen here, and the test passes.

I said all of this before, and maybe you also want to review the tests themselves, not just the console log.

> If this is an expected/correct exception, the testcase should trap it and report it correctly handled

It **is** correctly handled by the version class. What that class does it to log the exception to stdErr, but handle the problem gracefully, returning a version 0.0.0 in case the unthinkable happens that the resource file cannot be found or parsed correctly.

> rather than letting it output to console, right?

Wrong. The only way to suppress that test from logging to stdErr would be to redirect stdErr for the duration of the test, and like I also said before, this makes the test not thread-safe. Tests can be run in parallel. Please read what I talked about with Vladimir, all your questions are already answered, I am repeating myself here.

jkesselm:
Sorry -- until I saw it I didn't realize what that discussion was about.

OK.  I just wanted to check that what I was seeing was indeed what was expected. I'll do a full read-through of the code before commenting further. I expect to merge this barring something much more severe; it produces correct test results, and if a better solution turns up it can be refined later.

vlsi:
@kriegaex , right you are, the test will not be safe to run concurrently with the others, so you could add (`@Isolated`)[https://junit.org/junit5/docs/snapshot/api/org.junit.jupiter.api/org/junit/jupiter/api/parallel/Isolated.html] annotation to the test method so JUnit executes the test in isolation from the others. It does not hurt to have several isolated tests.

As you probably want to ensure the exception is logged, you probably want adding an assertion to the test.

So all-in-all it makes sense to capture the output (e.g. replace sytsem.err), assert that the output contains the expected patterns (e.g. "unable to pasre version", "xalan", ...), and mark the test as isolated.

Frankly, I do not understand why you automated virtually everything yet you insist that stderr must be eyeballed.

+100500 to what @jkesselm says, except "I'm probably going to come back and change it after the PR" which, in my opinion is doubtful. Of course, committers can commit code as they agree among the committers, but "dear contributor, say whatever you want in pr, but I will change it anyway" does not sound like a great practice overall. It is a tough call, and I would probably not add such a phrase.

I think @jkesselm knew the exception was expected in the test, and the comment https://github.com/apache/xalan-java/pull/133#issuecomment-1836632327 was a polite variation to double-check that and note that stacktraces in console for the expected failures are confusing.

jkesselm:
I see.... I think this is style more than substance, right now. I would have handled that exception differently in Version, which would have changed how the test behaved a bit. But I didn't write it, or write the test. This works, it's an improvement over prior state, it gets committed. 

Error-reporting architectural review can happen later when designing other unit tests and/or examining Xalan's progress/warning/error messages generally.

Still reading the code, but currently expect to merge.

kriegaex:
@vlsi: FWIW, @jkesselm just mentioned in a private chat that there seems to be some kind of internal (not 3rd party) logging framework in use in the project, despite me also having found ~700 occurrences of `System\.(out|err)\.print` in the code (excluding examples and ITs). I.e., it is not used universally. I would have expected only a handful of direct logs to stdOut and stdErr in framework/utility methods, and everything else that is supposed to log something or print something to the console in a CLI application to use utility methods. Then, it would have been easy to find those methods, used everywhere. But they are, in fact, not used everywhere. I guess, we both did not know that such logging infrastructure exists, otherwise probably you would have suggested it to me. then, I could have done something like
  * log a one-line warning to the info or warning channel,
  * log the full stack trace to the debug or error channel.
The test could then have used the appropriate log level and an in-memory logger and even inspect the logs from a unit test.

Joe could not tell me right away how to use this logging framework or where to copy code from, because he has just returned to the project after a long absence. But if it can be used outside a Xalan context (i.e. just from the version classes), in the future we could easily change the code to log there, even using multi-language resources according to Joe. That would, however, make the version classes do way more than the old ones did, but it would certainly be possible.

Because these are the very first unit tests in this project (not externalised like the ones in _xalan-test_), I did not want to create a precedent of a non-thread-safe test right in the first example. Like I said, it would be super easy to set `System.err` to a `PrintStream` wrapping a `ByteArrayOutputStream` for the duration of the test and even inspect the logging results. The result would be a silent test (no log output to the console), which would certainly be nice to have. But I guess, we should wait for Joe to explain to us how to use the internal log facilities, before we decide how to handle this in the future.

The cleanest thing would be to migrate the whole product away from internal tool classes and use 3rd party ones, specifically designed for that task. Less code to maintain in the long run. Same goes for the internal test framework. But both would be huge separate PRs, nothing to be debating withing the confines or a single, small PR like this one, which adds testing frameworks and a handful of initial tests.

kriegaex:
> so you could add [`@Isolated`](https://junit.org/junit5/docs/snapshot/api/org.junit.jupiter.api/org/junit/jupiter/api/parallel/Isolated.html) annotation to the test method so JUnit executes the test in isolation from the others. It does not hurt to have several isolated tests.

That would be an option. Like I said, I wanted to avoid it in the very first example. Bad precedent. But actually, it is not such a bad idea. Within the isolated class, the tests could still run in parallel, however, i.e. asserting on log output via a single redirected log stream would still be a problem. But the suppression problem would be worked around. The real solution is to log both using (not newly introducing) a framework on different log levels, like I said in my previous message.

PR #134, opened 2023-11-30
Delete superfluous doclet and taglet JARs

kriegaex:
@jkesselm


kriegaex:
> Yep. You've retained top-level tools/xalan2jdoc.jar, which is good

That was not my intention. I **have** deleted it and al so pushed the commit to the branch. For some reason, the PR was not updated on GitHub. Must be a glitchin their service, as usually that does not happen. Or you just merged before I pushed.

I absolutely wanted to delete it, as we agreed on in our chat. There is no need to keep it in the current version, as it is still in the Git history. I already turned it into source, but deleted the decompiled sources again, because they cannot work on JDK 8, as explained in my corresponding commit.

I am going to create a follow-up PR for the missing commit.

PR #137, opened 2023-12-03
Run version unit tests in isolation and assert on error logs

kriegaex:
Use JUnit Jupiter's `@Isolated`, because redirecting `System.err` is not thread-safe. If we have no logging framework with per-class log levels and the option to use in-memory loggers in place, the only way we can silence the test from logging stack traces and warnings to `System.err` is to redirect it to an in-memory stream and assert on its contents for verification.

I do not like this solution, but for the time being it is viable, albeit suboptimal.

@jkesselm, @vlsi, maybe this makes both of you a little happier on the expense of me being a little bit unhappier. But my heart is not in this project, so I can live with it for now.

### DIFF
--- a/xalan/src/test/java/org/apache/xalan/VersionTest.java
+++ b/xalan/src/test/java/org/apache/xalan/VersionTest.java
@@ -17,7 +17,11 @@
  */
 package org.apache.xalan;
 
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Isolated;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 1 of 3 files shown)


kriegaex:
OT: Code duplication got more, this time in the unit test classes. More and more, I am missing a _xalan-commons_ module...

xalan/src/test/java/org/apache/xalan/VersionTest.java
```diff
@@ -59,41 +85,47 @@ public void testCannotReadProperties() {
         .when(Version::getPropertiesStream)
         .thenThrow(NullPointerException.class);
       assertEquals("0.0.0", Version.readVersionNumber());
+      assertTrue(buffer.toString().contains("RuntimeException: Cannot read properties file"));
```
vlsi, line 88: `PrintStream` might buffer results so you probably want to add `mockPrintStream.flush()` before `buffer.toString()` (there are multiple uses of `buffer.toString()`, and it looks like all of them need adjustments

jkesselm:
If you can describe how a _commons module should be defined/bounded, or point me to such a guideline, we can at least put it into Jira as more technical debt. I presume we agree that the Maven build is already late enough and big enough that we don't want to add that complication until after cut-over.

vlsi:
`commons` discussion does not belong to this PR.

jkesselm:
Both of you: Take the argument about who provoked what elsewhere, please, and keep discussion here focused on technical issues related this PR. 

Where that line gets drawn is a judgement call, but I think we're drifting well out of that scope again.

If you're feeling a need for someone to have the last word in public, let it be me, OK?

jkesselm:
Looks like my own recent work introduced a merge conflict. My faulf for not having merged this sooner. I can handle that, I think.

PR #166, opened 2024-01-24
updated

maxfortun:
https://issues.apache.org/jira/browse/XALANJ-2725

### DIFF
--- a/serializer/src/main/java/org/apache/xml/serializer/ToStream.java
+++ b/serializer/src/main/java/org/apache/xml/serializer/ToStream.java
@@ -175,6 +175,10 @@ abstract public class ToStream extends SerializerBase
      */
     private boolean m_expandDTDEntities = true;
   
+    /**
+     * Track multibyte character in order to serialize when the whole byte sequence is available.
+     */
+    private char m_highUTF16Surrogate = 0;
 
     /**
      * Default constructor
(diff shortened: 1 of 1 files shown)


serializer/src/main/java/org/apache/xml/serializer/ToStream.java
```diff
@@ -1595,23 +1599,40 @@ else if (m_encodingInfo.isInEncoding(ch)) {
...
+                        // Store for later processing. We may be at the end of a buffer,
+						// and must wait till low surrogate arrives
+						// before we can do anything with this.
+                        writeOutCleanChars(chars, i, lastDirtyCharProcessed);
+                        m_highUTF16Surrogate = ch;
+                        lastDirtyCharProcessed = i;
+                    }
+                    else if (m_highUTF16Surrogate != 0 && Encodings.isLowUTF16Surrogate(ch)) {
+                        // The complete utf16 byte sequence is now available and may be serialized.
+                    	if (! m_encodingInfo.isInEncoding(m_highUTF16Surrogate, ch)) {
```
jkesselm, line 1612: I'm a bit confused by your intent here.
The high surrogate is never in any of the encodings, and should never be output by itself, so this case is firing incorrectly.

maxfortun: 1602: If ch is high surrogate, dump clean chars and retain ch. This works within the same buffer or across buffer boundaries
1610: if we have a retained high surrogate and ch is a valid low surrogate, we have 2 use cases. 
    1. encoding doesn't support the multibyte char, escape entity.
    2. encoding does support the multibyte char, output chars as is.

serializer/src/main/java/org/apache/xml/serializer/ToStream.java
```diff
@@ -1595,23 +1599,40 @@ else if (m_encodingInfo.isInEncoding(ch)) {
...
+						}
+                        lastDirtyCharProcessed = i;
+                        m_highUTF16Surrogate = 0;
                     }
                 	else {
                         // This is a fallback plan, we get here if the
                     	// encoding doesn't contain ch and it's not part
                     	// of a surrogate pair
                         // The right thing is to write out an entity
+						if(m_highUTF16Surrogate != 0) {
```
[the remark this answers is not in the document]
maxfortun: Btw, I did encounter this scenario, that's why I coded this. But in theory this should not be happening.

jkesselm:
Looks more plausible, though I'm still not sure writing out an NCE for an isolated high surrogate is a win.

I still sorta hate the idea of adding Still More Conditionals to the rendering loop, but I don't see a great alternative.

This doesn't seem to fix what slipped between 2419 and master, alas. Lemme work on that and get back to you.

jkesselm:
Getting back to this issue ... I've got an idea on how to handle the look-behind with a bit less overhead, essentially as inlined tail-recursion, so we can avoid testing for leftover high surrogate every time. Lemme see what I can do with that.

kubycsolutions:
I believe a fix for this issue has already been committed; if you agree, we should close this pull request.

PR #167, opened 2024-01-25
XALANJ-2618 Invalid encodings in Encodings.properties

cdamioli:
Encodings.properties contains invalig encodings.
Since Java 9+ it causes issues while serializing ISO-8859-1 characters

See https://issues.apache.org/jira/browse/XALANJ-2618 and https://issues.apache.org/jira/browse/XALANJ-2625


jkesselm:
We both have solutions (mine is at https://github.com/apache/xalan-java/tree/XALANJ-2618), I think I like my fix a bit better, since it resolves the problem without having to remove encoding names that, correct or not, someone might have been depending upon, and addresses the problem at a systemic level rather than just patching this one encoding set.

Any suggestions on the best way to make sure you get credit for isolating the issue while we check in my fix?

(I'm used to working in an environment where we cared more about the fix getting in than exactly who got credit for it, so my reflexes are not entirely right for public Git. I'm trying to retrain myself to be more sensitive about that, recognizing that contributor history is important for most folks.)

PR #186, opened 2024-04-18
XALANJ-2664: Determine the parent ClassLoader for ByteArrayClassLoader

jamesnetherton:


jkesselm:
The change looks harmless, but an explanation/justification would still be good practice, eg mention what breaks without the super(parent).

jamesnetherton:
@jkesselm there's an overview here:

https://issues.apache.org/jira/browse/XALANJ-2664

My use case is for XSLT with Apache Camel Quarkus where we support running applications natively on GraalVM. We have to do translet compilation at build time since it can't be done dynamically at runtime with GraalVM.

Due to how Quarkus organises its ClassLoaders for the build time augmentation phase, relying on the system ClassLoader is not guaranteed to work.

Hence using get `ObjectFactory` to figure out the parent ClassLoader. Which seems to be consistent with how things are done elsewhere, like in `TransletClassLoader`:

https://github.com/apache/xalan-java/blob/c8ddaade0e640383aead7c02a76759155922690e/xalan/src/main/java/org/apache/xalan/xsltc/trax/TemplatesImpl.java#L283-L288

https://github.com/apache/xalan-java/blob/c8ddaade0e640383aead7c02a76759155922690e/xalan/src/main/java/org/apache/xalan/xsltc/trax/TemplatesImpl.java#L119-L130

PR #191, opened 2024-11-26
Replace uses of StringBuffer with StringBuilder.

turbanoff:
StringBuffer is a legacy synchronized class. StringBuilder is a direct replacement to StringBuffer which generally have better performance. It's quite important for performance in fresh versions of Java which disabled/removed biased locking.

### DIFF
--- a/xalan/src/main/java/org/apache/xalan/processor/XSLTAttributeDef.java
+++ b/xalan/src/main/java/org/apache/xalan/processor/XSLTAttributeDef.java
@@ -456,7 +456,7 @@ public class XSLTAttributeDef
         return m_setterString;
       }
 
-      StringBuffer outBuf = new StringBuffer();
+      StringBuilder outBuf = new StringBuilder();
 
       outBuf.append("set");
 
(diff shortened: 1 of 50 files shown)


elharo (commented):
This PR might be correct, but FYI  StringBuilder is *not* a direct replacement for StringBuffer. It's fine for local variables that don't escape method scope. Other usages need careful consideration of whether the synchronization properties are important.

xalan/src/main/java/org/apache/xalan/processor/XSLTAttributeDef.java
```diff
@@ -1532,12 +1532,12 @@ else if (cl == Long.class)
...
   /**
-   * StringBuffer containing comma delimited list of valid values for ENUM type.
+   * StringBuilder containing comma delimited list of valid values for ENUM type.
    * Used to build error message.
    */
-  private StringBuffer getListOfEnums() 
+  private StringBuilder getListOfEnums()
   {
-     StringBuffer enumNamesList = new StringBuffer();            
+      StringBuilder enumNamesList = new StringBuilder();
```
elharo, line 1540: indentation looks off by one space

PR #194, opened 2025-09-17
Valhalla JEP 401 compatibility

ericcaspole:
This small change appears to bring xalan into compliance with JEP 390[1], which was a warning in preparation for JEP 401[2] by switching synchronized Integers to Objects,  where the value of the Integer is never used, and this allows xalan dependent benchmarks, such as Dacapo xalan and fop, to run under builds of the Valhalla project[3] with --enable-preview. 

[1] https://openjdk.org/jeps/390
[2] https://openjdk.org/jeps/401
[3] https://github.com/openjdk/valhalla

### DIFF
--- a/serializer/src/main/java/org/apache/xml/serializer/OutputPropertiesFactory.java
+++ b/serializer/src/main/java/org/apache/xml/serializer/OutputPropertiesFactory.java
@@ -183,7 +183,7 @@ public final class OutputPropertiesFactory
     private static final int S_XALAN_PREFIX_LEN = S_XALAN_PREFIX.length();
 
     /** Synchronization object for lazy initialization of the above tables. */
-    private static Integer m_synch_object = new Integer(1);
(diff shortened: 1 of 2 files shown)


Yqwed:
> On the other hand, if we are using an object only as a semaphore anchor, we honestly don't care what its type is.

Exactly. There is absolutely no reason to prefer `new Integer(1)` or `new Boolean(true)` over `new Object()` if it is used for synchronization only.

>  I am currently inclined to classify it as "at worst, harmless."

I'd disagree about harmless - as it is xalan can't be used in early access valhalla builds. Also OpenJDK team plans to make it a preview feature in JDK 27 (to be released this autumn) and xalan users will need to patch this library.

This change has zero risks and only helps Java ecosystem as a whole.

Yqwed:
Thank you!

liach:
Thank you!

PR #195, opened 2025-09-25
Set Automatic-Module-Name to org.apache.xalan

SingingBush:
As this project is still supporting JDK 8 for now I've just used the maven jar plugin to set *org.apache.xalan* as the Automatic-Module-Name. In the future Xalan should have a module-info.java for full JPMS support

### DIFF
--- a/xalan/pom.xml
+++ b/xalan/pom.xml
@@ -34,6 +34,17 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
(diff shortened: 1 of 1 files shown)


SingingBush:
@mukulga it seems that all current development activity is in the [xalan-j_xslt3.0_mvn](https://github.com/apache/xalan-java/tree/xalan-j_xslt3.0_mvn) branch but it doesn't define an Automatic-Module-Name in the pom.

please can the equivalent of this PR be done on that branch as well so that `<Automatic-Module-Name>org.apache.xalan</Automatic-Module-Name>` is in the next release.

